PR #45, opened 2018-08-04
Update classifier.md

vikaskyadav:


jhfjhfj1:
The classifier.md is automatically generated from the docstrings in the code.
You may want to modify the docstrings instead.

Thank you for your contribution.

PR #54, opened 2018-08-06
Qq

jhfjhfj1:


jhfjhfj1:
@song3134 Please add some pytest to make sure the coverage from coveralls doesn't drop.

Issue #91, opened 2018-08-16
overflow encountered in exp on bayesian.py line 89


tl-yang:
### Bug Description
When running image_classifier. If we use a big MAX_MODEL_NUM (1000). https://github.com/jhfjhfj1/autokeras/blob/master/autokeras/bayesian.py#L89 would throw an overflow warning. Which causes a FileNotFound Error to be raised.

Here is the stack trace:

```
  File "/home/tsunglin/autokeras/autokeras/image_classifier.py", line 229, in fit
    run_searcher_once(train_data, test_data, self.path, int(time_remain))
  File "/home/tsunglin/autokeras/autokeras/image_classifier.py", line 42, in run_searcher_once
    searcher.search(train_data, test_data, timeout)
  File "/home/tsunglin/autokeras/autokeras/search.py", line 189, in search
    new_graph, new_father_id = self.maximize_acq(timeout)
  File "/home/tsunglin/autokeras/autokeras/search.py", line 268, in maximize_acq
    nm_graph = self.load_model_by_id(father_id)
  File "/home/tsunglin/autokeras/autokeras/search.py", line 109, in load_model_by_id
    return pickle_from_file(os.path.join(self.path, str(model_id) + '.h5'))
  File "/home/tsunglin/autokeras/autokeras/utils.py", line 190, in pickle_from_file
    return pickle.load(open(path, 'rb'))
FileNotFoundError: [Errno 2] No such file or directory: 'tests/resources/temp/None.h5'
```

### Reproducing Steps
  * Step 1: open test_image_classifier
  * Step 2: change Constant.MAX_MODEL_NUM to 1000 in test_save_continue function.
  * Step 3: run the unit test and you'll see the exception.

### Expected Behavior
Without error

### Setup Details
Include the details about the versions of:
 - OS type and version: Ubuntu 16.04
 - Python: 3.6
 - autokeras: commit f78681805241bd017d75c9c24844e82a051dae9a
 - scikit-learn: 0.19.1
 - numpy:1.14.5
 - keras: 2.2.2
 - scipy: 
 - tensorflow: 1.8.0
 - pytorch: 0.4.0

PR #96, opened 2018-08-17
[MRG] Test timeout resume

tl-yang:
This pull request adds a test for the timeout resume function.

- This test will sometimes throw a FileNotFound exception. Blocking by #91
- Warning message:  
image

- This timeout in this unit test should be able to finish executing optimize_acq, but not enough to find all(1000) models


tests/test_image_supervised.py
```diff
@@ -69,11 +69,43 @@ def test_timout(_):
...
 
 
+@patch('multiprocessing.Pool', new=MockProcess)
+@patch('autokeras.search.ModelTrainer.train_model', side_effect=mock_train)
+def test_timout_resume(_):
+    Constant.MAX_ITER_NUM = 1
+    # make it impossible to complete within 10sec
+    Constant.MAX_MODEL_NUM = 1000
+    Constant.SEARCH_MAX_ITER = 1
+    Constant.T_MIN = 0.8
```
jhfjhfj1: @tl-yang 
If the timeout is working, it doesn't need to use T_MIN=0.8 to reduce the number of models searched during optmize acq. This line is the reason for the FileNotFound exception. Please try to delete it and see if it still works.
Thanks.

PR #101, opened 2018-08-20
Update classifier.md

parinaya-007:
In classifier.md file, the headings were not in the correct format as per Markdown. So corrected that.


jhfjhfj1:
@parinaya-007 Thank you for your contribution.
However, our Markdown files are generated by executing mkdocs/autogen.py.
So it is the file that should be changed. You may change it in another merge request.

I will remove the duplication to avoid this in the future.
Thanks.

PR #170, opened 2018-09-03
doc string changed for augment

swenkel:
I added proper documentation for class ImageSupervised  arg 'augment'. It is 'None' by default. However, if it is 'None', then it uses Constant.DATA_AUGMENTATION which is 'True'. This is misleading when trying things out.


jhfjhfj1:
@swenkel Thank you so much for the contribution!

PR #173, opened 2018-09-04
Fixed load.py and lifted it from the zip file.

### LINKED ISSUE #152: load_raw_image.zip example is broken
### Bug Description
The load_raw_image example in the load_raw_image.zip under examples still contains the old naming convention:

```from autokeras.classifier import load_image_dataset```

While the current (and working) example in the github pages reads:

```from autokeras.image_supervised import load_image_dataset```

### Reproducing Steps
  * Step 1: Download zip and extract
  * Step 2: See for yourself

### Expected Behavior
A working example on download :-)

Might I suggest removing the code from the zip file and just placing it on git. So it can more easily be spotted and updated. Since I assume the image files will remain the same leaving them in a zip makes sense I guess.

dietercastel:
This pull request resolves #152.

I extracted load.py from the zip and applied the aforementioned fix to. I added some extra code to make the example run a classifier on the loaded images.

### DIFF
--- a/examples/load_raw_image/load.py
+++ b/examples/load_raw_image/load.py
@@ -1,0 +1,31 @@
+#To run first unzip ../load_raw_image_data.zip into ../
+# so the train and test directories reside in this directory 
+
+from autokeras.image_supervised import load_image_dataset
+from keras.datasets import mnist
+from autokeras import ImageClassifier
+
+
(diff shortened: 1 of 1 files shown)


jhfjhfj1:
@dietercastel Thank you for the contribution.

PR #202, opened 2018-09-18
[MRG] First-stab working Image1DClassified (#88)

### LINKED ISSUE #88: ImageClassifier1D
### Feature Description

CNN structure is becoming more and more widely used in text classification. Unlike image in shape [height, weight, channel], in this case the input will be in shape [1, n, embedding size], where n is the length of the text. Currently such input will cause error in autokeras, since "1 - filter_size" will be less than zero. So I think autokeras should automatically detect such input shape and then apply 1D architecture search like 1D convolution, 1D pooling, etc. 

### Reason

Autokeras now is mainly for 2D input like image. It could be more flexible.

### Solution

Add some codes to detect the shape of the input, and to search specific network structures optimized for 1D, 2D or 3D tensor depending on the shape of the input.

wernergkrebs:
This pull request resolves #88.

Baseline, working first stab at 1D classifier for autokeras re-using much of the existing autokeras functionality. Included is an example, pytest files, and documentation in comments, with low-hanging fruit suggestions for further improvements. It does not work as well as I'd like (auto-generated topologies are not optimal for 1D problems, and typically do far better with manual topologies, but I'm a data scientist....) However, it is a working first-stab at autokeras functionality for auto-generating 1D numerical classifiers. (It's suggested that further extending this code by auto-subsituting numeric labels for text strings would then result in a first-stab generalized classifier.) Some further tweaks to the Bayesian topology generated to optimize it for 1D rather than 2D problems should greatly improve this code. 

Hopefully this code and some of the comment suggestions will inspire someone to add a more full-featured, better-optimized 1D classifier to autokeras.


jhfjhfj1:
@wernergkrebs Thank you so much for the contribution!
Would you please fix the CI so that we can start a code review?

jhfjhfj1:
@wernergkrebs Usually for our unit test, it can finish within 1 or two minutes.
There must be something wrong with your test code.
Might because of the mock is not correct.

wernergkrebs:
> @wernergkrebs Usually for our unit test, it can finish within 1 or two minutes.
> There must be something wrong with your test code.
> Might because of the mock is not correct.

HI @jhfjhfj1 , I'm not going to argue with you on this. As project owner, you will often be called upon to diagnose issues with your code, and should not expect others to do this for you.

In my tests, your code passed tests but did so slowly, and the output from the GitHub build in fact shows it timing out in your existing code sections, which were not touched. You are using a random number generator to generate test cases, so variable build and test times are to be expected with such a setup.

My assertion is that the tests pass, they just sometimes take slightly then the allowed GitHub timeout, and the length of time for the tests to pass is variable due to the use of random number generator to generate test cases. I believe the issue can be mitigated by following Github's own suggestion to either increase test timeout length or request GitHub retry the build. 

In any event, if new tests are added, one would expect test builds to take longer, so running into a timeout as code grows is not unexpected. It doesn't seem like you've investigated whether this is possible, or tried to run the tests in your own build environment to get timings on how long the tests take to pass on average, but simply have asserted without evidence the problem "must be" in the newly checked in code, without examining the issue at all or trying to do your own test builds, which only takes a few minutes.

jhfjhfj1:
@wernergkrebs 

Thank you for the contribution and suggestions!
You are right! It works now.

I will improve on how I manage these pull requests, and hope to see more comments and contributions from you in the future.

Would you please increase the test coverage a little bit for your code?
https://coveralls.io/builds/19272393/source?filename=autokeras%2Fimage1D_supervised.py#L139
Then I will merge it.

Thanks.

jhfjhfj1:
Since we have found a more thorogh solution for this task, we close the pr.
Thank you so much for your contribution.

PR #238, opened 2018-10-04
[MRG]Fit param. Resolve issue #209

### LINKED ISSUE #209: Specify the validation set in the parameters of fit()
### Feature Description
In supervised.py:

`fit(self, x_train=None, y_train=None, time_limit=None):`
to
`fit(self, x_train=None, y_train=None, x_test=None, y_test=None, time_limit=None):`

And then update the implemented task modules to use the test subset (if provided) instead of splitting the train.

### Reason

In some cases the automatic split of 80:20 (in the case of ImageSupervised) is not good enough and the model has to be evaluated on a specific subset which is already available.

### Solution
I can submit a pull request if you consider this as a needed feature. I'd argue that it doesn't increase the complexity of using the project in its simplest form since the old automatic splitting is also preserved.

### Alternative Solutions
Leave it as it is.

tl-yang:
This pull request resolves #209 .

- Remove x_train, y_train default param, add x_test, y_test to default param.

- Add .coveragerc to exclude abstract function.


autokeras/image_supervised.py
```diff
@@ -186,30 +186,24 @@ def fit(self, x_train=None, y_train=None, time_limit=None):
...
-
+        if x_test is None or y_test is None:
+            # Divide training data into training and testing data.
+            x_train, x_test, y_train, y_test = train_test_split(x_train, y_train,
+                                                                test_size=min(Constant.VALIDATION_SET_SIZE,
+                                                                              int(len(y_train) * 0.2)),
+                                                                random_state=42)
         # Transform x_train
         if self.data_transformer is None:
             self.data_transformer = DataTransformer(x_train, augment=self.augment)
```
jhfjhfj1: The data transformer should be initialized before the split. The transformer init calculates the mean and standard deviation from the data. It should use the whole training set.

PR #295, opened 2018-11-05
[WIP] 1D3D image classification and regression

### LINKED ISSUE #84: ImageClassifier3D
### Feature Description
 
Your library can do 3D convolutional auto keras.

### Reason

Input images are 3D images, such as 3D CT scan images.  

### Solution

If you add this feature, it will benefit many people.
 
### Benchmark Datasets
[3D-MINST](https://www.kaggle.com/daavoo/3d-mnist/data#full_dataset_vectors.h5)

### LINKED ISSUE #88: ImageClassifier1D

jhfjhfj1:
This pull request resolves #88, resolves #84.

### DIFF
--- a/autokeras/cnn_module.py
+++ b/autokeras/cnn_module.py
@@ -7,11 +7,6 @@ from autokeras.search import Searcher, train
 from autokeras.utils import pickle_from_file
 
 
-def _run_searcher_once(train_data, test_data, path, timeout):
-    if Constant.LIMIT_MEMORY:
-        pass
-    searcher = pickle_from_file(os.path.join(path, 'searcher'))
-    searcher.search(train_data, test_data, timeout)
 
 
 class CnnModule(object):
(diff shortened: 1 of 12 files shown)


mstern2493:
@haifeng-jin , thanks so much for doing this! Is 1d image classification (with multiple channels) still supported? If so, what does the usage look like? I have been trying to get an input of shape [batch_size x channels x num_samples] to work where num_samples is the dimension of the 1d image, but the ak.ImageInput seems to treat it as 2d image (using the channels) and something like a ResNetBlock won't work with an ak.Input of that shape. Ideally, I'd like to run the input through both a ResNet and an RNN, but am having difficulties reconciling that with the documented tutorials.

PR #383, opened 2018-12-17
[MRG]Single process Search #99

### LINKED ISSUE #99: Incompatible with Google Colaboratory
### Bug Description
As stated [here](https://stackoverflow.com/questions/51835926/google-colab-runtime-dies-when-using-autokeras), Autokeras kills the Google Colaboratory kernel, in CPU or GPU mode.

### Reproducing Steps
Run the following [notebook](https://colab.research.google.com/drive/1REPj8pVSpSUvvufHBY693RmkXjnBSKcg), which is a simple copy-paste of the AutoKeras tutorial, with a slight modification: in the AutoKeras API, the package autokeras.image_classifier seems to have been renamed to autokeras.image_supervised, and the [documentation](https://autokeras.com/start/#example) is not yet up to date.

### Expected Behavior
It should run, since Colaboratory runs Python 3.6

### Additional context
Colaboratory seems incompatible with the Python multiprocessing API. 

I modified the code of search.py/Searcher in order to remove it, and it worked.
More clearly, I replaced this:
` multiprocessing.set_start_method('spawn', force=True)`
`pool = multiprocessing.Pool(1)`
`train_results = pool.map_async(train, [(graph, train_data, test_data, self.trainer_args,
                                                os.path.join(self.path, str(model_id) + '.png'),
                                                self.metric, self.loss, self.verbose)])`        
By this:
`
        train_results = train((graph, train_data, test_data, self.trainer_args,
                                                os.path.join(self.path, str(model_id) + '.png'),
                                                self.metric, self.loss, self.verbose))
`

Maybe AutoKeras could have an option in order to desactivate the use of that API ?

tl-yang:
This pull request resolves #99.

- Move timeout to model trainer
- Create Single process option for searcher


autokeras/constant.py
```diff
@@ -37,7 +37,7 @@ class Constant:
     MAX_NO_IMPROVEMENT_NUM = 5
     MAX_BATCH_SIZE = 128
     LIMIT_MEMORY = False
-    SEARCH_MAX_ITER = 200
+    SEARCH_MAX_ITER = 1
```
jhfjhfj1: Why this line is changed to 1?

autokeras/nn/layers.py
```diff
@@ -300,6 +300,7 @@ def __init__(self,
         super().__init__(input_node, output_node)
         self.kernel_size = kernel_size if kernel_size is not None else Constant.POOLING_KERNEL_SIZE
         self.stride = stride if stride is not None else self.kernel_size
+        # TODO
```
jhfjhfj1: What is this to do?

autokeras/nn/model_trainer.py
```diff
@@ -33,6 +36,7 @@ def __init__(self,
                  loss_function,
                  train_data,
                  test_data=None,
+                 timeout=0,
```
jhfjhfj1: 0 here may not be safe. Can we change it to none?

autokeras/constant.py
tl-yang: It must be an accident.

autokeras/nn/layers.py
tl-yang: for padding, i think. But since we've added the padding, I'll remove it

autokeras/nn/model_trainer.py
tl-yang: Sure, I'll use none to represent the no timeout

tl-yang:
@jhfjhfj1  I found several bug in this pr, might postpone a little bit. Sorry

tl-yang:
> @jhfjhfj1 I found several bug in this pr, might postpone a little bit. Sorry

@jhfjhfj1 I think this version should be better, please let me know if there's any issue.

autokeras/search.py
```diff
@@ -187,34 +190,40 @@ def search(self, train_data, test_data, timeout=60 * 60 * 24):
...
-                self.descriptors.append(generated_graph.extract_descriptor())
-
-            remaining_time = timeout - (time.time() - start_time)
-            if remaining_time <= 0:
-                raise TimeoutError
-            metric_value, loss, graph = q.get(timeout=remaining_time)
-
-            if self.verbose and searched:
+            generated_other_info, generated_graph, new_model_id = self._search_common(q)
+            #
```
jhfjhfj1: Why not just delete this part?

PR #395, opened 2018-12-25
[WIP]Add basic multiple gpu support

### LINKED ISSUE #60: Parallel on Multiple GPUs
I tried fitting the data on amazon p2.8xlarge instance(which has 8 GPUs), but it just fills the VRAM of one GPU and crashes with the cuda out of memory error. Is there any way of using all 8 GPUs?

tl-yang:
This pull request resolves #60.

- Support parallel on multiple GPU
- get_device() now only determine GPU or CPU.
- Cannot disable multiple GPU training for now, users can only constraint the gpu usage by specifying CUDA_VISIBLE_DEVICES environment variable


tl-yang:
@jhfjhfj1 I think this function is ready to merge. Could you please comment on whether you like the idea of using multi_gpu as a default behavior (can be constraint by CUDA_VISIBLE_DEVICE environment variable) or adding one more parameter to the ModelTrainer is more preferable?

jhfjhfj1:
@tl-yang I think it is fine to use multi-gpu by default. So does using env variable means we don't use GPUtil anymore? If we still use GPUtil, how and why should we use it?

I think it would be a good idea that we don't use GPUtil, but let the users to use GPUtil to set the env var to constrain the GPU usage of autokeras.

tl-yang:
@jhfjhfj1 GPUtil is used in text_preprocessed.py. Not by utils.py. So I think we need to confirm with @boyuangong ?

jhfjhfj1:
@tl-yang I will merge this pull request. We can discuss with Boyuan later. Thanks.

PR #404, opened 2018-12-31
Voice generate pre [MRG]

### LINKED ISSUE #136: SpeechSynthesizer
### Suggested Name

SpeechSynthesizer

### Task Description

The input is a text file in English. The output is an audio reading the text content aloud.

### Evaluation Metrics

Human Evaluation

### Benchmark Datasets

[VCTK](http://homepages.inf.ed.ac.uk/jyamagis/page3/page58/page58.html)

boyuangong:
This pull request resolves #136 .


boyuangong:
The test case and example will cause weird error.

autokeras/__init__.py
```diff
@@ -4,5 +4,5 @@
 
 from autokeras.net_module import CnnModule, MlpModule
 
-from autokeras.pretrained.object_detector import ObjectDetector
-from autokeras.pretrained.face_detector import FaceDetector
+# from autokeras.pretrained.object_detector import ObjectDetector
```
jhfjhfj1: We need these two lines.
The user should be able to directly import the pretrained model from autokeras.something.

boyuangong: I moveed this to pretrained __init__.py . My feeling is that the pretrained should be independent to other functions. When load current pretrain model, one can use autokeras.pretrained.something.

PR #477, opened 2019-01-22
Syntax error resolved

Mainak431:
This solved a syntax error in utils.py


jhfjhfj1:
@Mainak431 It is a feature of Python3 instead of a syntax error.
Thanks.

PR #479, opened 2019-01-23
[MRG] Support multicore/multiprocess for read_images function

### LINKED ISSUE #478: feature_request: Support multicore/multiprocess for read_images function
### Feature Description
Support multicore for `read_images` function in `image/image_supervised.py`.

### Reason
As [read_images function in image_supervised.py](https://github.com/jhfjhfj1/autokeras/blob/36752fcbce58bd1a26c17144637cc0ecbef83da0/autokeras/image/image_supervised.py#L14) does not support multi-core(just read files one by one), so it's **too slow** when reading image files more than 10k files.

### Solution
Use multiprocessing or equivalent one in `read_images` function as a default, or optional.

Beomi:
This pull request resolves #478 

- Add `_image_to_array` function to read each images.
- Use `multiprocessing.Pool` and `multiprocessing.cpu_count` to use multicore fully.

### DIFF
--- a/autokeras/image/image_supervised.py
+++ b/autokeras/image/image_supervised.py
@@ -1,6 +1,7 @@
 import os
 from abc import ABC
 import numpy as np
+from multiprocessing import Pool, cpu_count
 
 from autokeras.constant import Constant
 from autokeras.nn.loss_function import classification_loss, regression_loss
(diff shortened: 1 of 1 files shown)


jhfjhfj1:
@Beomi Thank you for your contribution. This is a very nice feature!
However, you need to add one more thing.
A default parameter in the function call to disable the parallel.
def read_images(img_file_names, images_dir_path, parallel=True).
Since many platforms don't support multiprocessing, the program would crash.
Remember to change the docstrings, too.

Thank you!

Beomi:
Got it. I made the `parallel` argument as an optional one.
And I just found there are other functions that could be parallelized, `compute_image_resize_params` and `resize_image_data`.
@jhfjhfj1, How do you think about parallelizing those functions ?

Hsiny:
I really hope resize_image_data() can be parallelized. It tasks a long time to preprocessing.I have 10 cpu cores,while the cpu usage is only 10%.

PR #535, opened 2019-02-12
[MRG] Implementing text classifier with BERT

satyakesav:

Cleaning up unnecessary code or functionality for implementing text classification.


satyakesav:
> This pull request looks good.
> However, I prefer to merge it after you finish this module.
> According to github-flow, the master branch should always be deployable.
> If I merge now, it will prevent us from releasing any new versions during your development of this module.
> 
> If we don't merge for now, I don't think anyone else would change the code of this module to create any conflict with yours.
> 
> Thanks.

Ok. I preferred to do it in multiple commits since the changes are very big and it will be easy for both code review and pushing the code. But, as you pointed about the rule, I will address the complete changes in this branch.

satyakesav:
> The coverall seems broke. I cannot see the test coverage. What is the coverage now?

Around 90.3% when I last saw. There is a decrease of 3%. Is that fine? It's heavily experimental to see how to mock BERT functionality at lower level to improve on the coverage but not sure what gain we might get. I will follow upon that as well for a later PR, if that's okay?

jhfjhfj1:
> > The coverall seems broke. I cannot see the test coverage. What is the coverage now?
> 
> Around 90.3% when I last saw. There is a decrease of 3%. Is that fine? It's heavily experimental to see how to mock BERT functionality at lower level to improve on the coverage but not sure what gain we might get. I will follow upon that as well for a later PR, if that's okay?

A decrease of 3% is too much. It usually means a large part of the code is not covered. You need to look into the coverage report locally, to see which part is missing. We can schedule a meeting to review the code together.

PR #569, opened 2019-03-06
check ipython kernel variant  to avoid multiprocess issue when running in ipython kernel

ghmole:
the Searcher.mp_search() function will throw exception when running in ipython kernel

just modify this to make debug run in ipthon kernel


tl-yang:
But I'm wondering what ipython you're running? I tried it in Jupyter notebook, it works without issue?

ghmole:
I run it in anaconda spyder, the mp issue always happens



sleepmole@126.com
 
From: tl-yang
Date: 2019-03-07 01:52
To: jhfjhfj1/autokeras
CC: ghmole; Author
Subject: Re: [jhfjhfj1/autokeras] check ipython kernel variant to avoid multiprocess issue when running in ipython kernel (#569)
But I'm wondering what ipython you're running? I tried it in Jupyter notebook, it works without issue?

tl-yang:
I think we've also encountered some other issues with spyder @jhfjhfj1 do you think we should support spyder environmen?

PR #584, opened 2019-03-16
[MRG] N-scalar output for ImageRegressor Class

BDANG:
This pull request resolves #ISSUE_NUMBER.

ImageRegressor was modified to support N output nodes. I found that ImageRegressor's output shape (examples, 1) to be too restrictive.

The modified class supports output shapes of (examples, N). For example, you could train models to crop images; the output would be (examples, 4) where the 4 scalars represent the top, right, bottom, left crop distances.

### DIFF
--- a/autokeras/image/image_supervised.py
+++ b/autokeras/image/image_supervised.py
@@ -216,7 +216,7 @@ class ImageClassifier(ImageSupervised):
         return self.y_encoder.n_classes
 
     def export_autokeras_model(self, model_file_name):
-        """ Creates and Exports the AutoKeras model to the given filename.
+        """Creates and Exports the AutoKeras model to the given filename.
         
         Args:
             model_file_name: A string containing the name of the file to which the model should be saved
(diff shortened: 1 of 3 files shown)


jhfjhfj1:
@BDANG Thank you very much for your contribution!
I have 2 questions for this PR.
1. Is it possible to make this class the same class as ImageRegressor, but make the regressor adaptively choose the "N" (may equal to 1 or not) according to the input data?
2. Would you refer us to any open dataset that has such kind of "N Regression" requirements?

BDANG:
@jhfjhfj1 

1) It's entirely possible to modify ImageRegressor, yeah.

2) Unfortunately, I'm not aware of an open datasets that apply N-regression. But if we take [NVIDIA's research on modelling steering angle](https://arxiv.org/pdf/1604.07316v1.pdf), You could forecast maybe multiple steering angles (immediate steering angle and future steering angle).

If you want me to proceed with modifying ImageRegressor, I'm happy to do so!

jhfjhfj1:
@BDANG Since we don't have any available open datasets on the NRegressor, I have concerns about the users need on this task.
So it would be great if you can just make the current Regressor capable of such multi-column regression by setting the default parameter instead of creating a new class.
In this way, we can support this functionality and add no extra maintenance burden to the project.
Thank you!

BDANG:
@jhfjhfj1 ready for review!

PR #646, opened 2019-06-04
Refactor

jhfjhfj1:
Addressed the code review comments.
For the "trails" parameter, the user should have a way to control the search time or budget briefly if they want.
We can temporarily remove it and figure out a new way later.

The comments are pasted below for convenience.

all files:

private things should be private: attributes and methods that aren't meant to be accessed by users should be prefixed with `_`.
the conventional prefix for a number in TF is `num_`, not `n_`
Classes that don't have a parent class should inherit `object`

auto_model.py:

AutoModel should inherit from HyperModel
mechanisms for keeping track of state (`trails`) should be documented and should be moved out of the public API
Configurability of the Tuner should be moved out of the public API. In fact we will probably need to package tuning algorithms directly on each AutoModel in the future.

hyper_block.py:

we can remove `build_output` and rename it to `build`
we only need the following blocks:
DenseBlock
SequenceBlock
RNNBlock
ImageBlock
ResNetBlock
XceptionBlock
Merge, Flatten, RepeatVector, TimeDistributed
HyperGraph -> rename GraphAutoModel, subclass of AutoModel -- the Functional API would be one way of instantiating a kind of AutoModel
ClassificationHead, RegressionHead: we need constructor arguments to specify the number of classes and the shape of the targets.
we don't need separate classes for different target shapes (we can remove the other 2 classes)

hyper_node.py:

Just Node, it's cleaner
We will need other inputs than ImageInput and Input (TextInput)


fchollet:
Also we need to discuss more about `HierarchicalHyperParameters` and how to handle name scopes.

jhfjhfj1:
Several things to change:
1. Remove the compile() function of AutoModel.
The heads should handle the loss and metrics (use fixed value for now). The HyperModel should handle the optimizer (make it a tunable variable).

2. The handling of name_scope should be done in the base classes, i.e., HyperModel, HyperParameters.
We can delete HierarchicalHyperParameters.
Use the shadowing mechanism to wrap the build function in the HyperModel class.
In the retrieving function of HyperParameters, it should include the name_scope.

3. We let the user define "trial".
4. We don't let the user specify "tuner".

autokeras/auto/auto_pipeline.py
```diff
@@ -1,16 +1,25 @@
-class AutoPipeline:
+from autokeras import AutoModel
+
+
+class AutoPipeline(AutoModel):
```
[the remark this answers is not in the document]
jhfjhfj1: Yes. I will add the classes. ImageClassifier should be one of them.

jhfjhfj1:
I am still working on the name_scope part. Other comments has been addressed. @fchollet

jhfjhfj1:
The name_scope is finished. Ready for another review. Thanks.@fchollet

autokeras/auto/auto_pipeline.py
```diff
@@ -1,29 +1,30 @@
...
-
-        Arguments:
-            x_train: An instance of numpy.ndarray.
-            y_train: An instance of numpy.ndarray.
-        """
+    def fit(self, **kwargs):
+        """Tuning the model. """
         pass
 
     def predict(self, x_test, postprocessing=True):
```
fchollet: just `x`

autokeras/auto/auto_pipeline.py
```diff
@@ -1,29 +1,30 @@
...
         """Predict the output for a given testing data.
 
         Arguments:
-            x_test: The x_test should be a numpy.ndarray.
+            x_test: An instance compatible for input to a Keras Model.
```
fchollet: "Data in a format that can be passed to a Keras model: either a Numpy array, a Python generator of arrays, or a TensorFlow Dataset.`

autokeras/auto/auto_model.py
```diff
@@ -18,59 +22,43 @@ class AutoModel(object):
...
+
         # Prepare the dataset
         if validation_data is None:
-            (x, y), (x_val, y_val) = split_train_to_valid(x, y)
+            (x, y), (x_val, y_val) = layer_utils.split_train_to_valid(x, y)
             validation_data = x_val, y_val
 
         # TODO: allow early stop if epochs is not specified.
-        self.tuner.search(trails,
+        self.tuner.search(num_trials=num_trials,
```
fchollet: I would recommend just naming this `trials` (like we have `epochs` or `steps`, for instance). The `num_` prefix is only useful if there is some potential confusion about the meaning of the name. For instance, if you have an `int` argument `layers`, that is confusing (it could be a list of layer instances?), so better to use `num_layers` in that case.

But for `trials` I think it is unambiguous that it is a number.

PR #650, opened 2019-06-07
[MRG] Simple version Image Pipeline Implemented

jhfjhfj1:
### Pull Request Status 
MRG

Implemented a simple version of ImageBlock and used it for the ImageClassifier and ImageRegressor.


autokeras/auto/image.py
```diff
@@ -0,0 +1,130 @@
...
+        new_space = hyperparameters.space[:]
+        if not self.allow_new_parameters and set(old_space) != set(new_space):
+            diff = set(new_space) - set(old_space)
+            raise RuntimeError(
+                'The hypermodel has requested a parameter that was not part '
+                'of `hyperparameters`, '
+                'yet `allow_new_parameters` is set to False. '
+                'The unknown parameters are: {diff}'.format(diff=diff))
+
+        datagen = tf.keras.preprocessing.image.ImageDataGenerator(
```
fchollet: Prefer using tf.data for a pipeline. Reference: https://www.tensorflow.org/tutorials/load_data/images

You'll probably need some basic utility for data augmentation with tf.data, let's touch base on that later.

autokeras/auto/image.py
```diff
@@ -0,0 +1,130 @@
...
+from autokeras import tuner
+
+
+class ImageTuner(tuner.SequentialRandomSearch):
+
+    def _run(self, hyperparameters, fit_kwargs):
+        # Build a model instance.
+        model = self.hypermodel.build(hyperparameters)
+
+        # Optionally disallow hyperparameters defined on the fly.
```
fchollet: This block should be abstracted as a method on the base class.

autokeras/auto/image.py
```diff
@@ -0,0 +1,130 @@
...
+            # randomly flip images
+            horizontal_flip=True,
+            # randomly flip images
+            vertical_flip=False)
+
+        datagen.fit(fit_kwargs['x'])
+        fit_kwargs['batch_size'] = min(len(fit_kwargs['x']),
+                                       fit_kwargs.get(
+                                           'batch_size',
+                                           const.Constant.BATCH_SIZE))
```
fchollet: The batch size should be at least configurable (e.g. in `__init__`)

autokeras/auto/image.py
```diff
@@ -0,0 +1,130 @@
...
+        return model, feedback
+
+
+class ImageSupervised(auto_pipeline.AutoPipeline):
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.image_block = hyper_block.ImageBlock()
+        self.head = None
+        self.x_normalizer = processor.Normalizer()
```
fchollet: just `normalizer` or `input_normalizer`, in general avoid single-letter notations

autokeras/auto/image.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def fit(self, x=None, y=None, **kwargs):
+        self.tuner = ImageTuner(self, metrics=self.head.metrics)
+        self.x_normalizer.fit(x)
+        self.inputs = [hyper_node.ImageInput()]
+        super().fit(x=self.x_normalizer.transform(x), y=y, **kwargs)
```
fchollet: In the future `normalizer` will be a preprocessing layer part of the model (added in `build`).

autokeras/auto/image.py
```diff
@@ -0,0 +1,130 @@
...
+
+        # Train model
+        history = model.fit_generator(data_flow, **temp_fit_kwargs)
+
+        metric_name = model.metrics_names[1]
+        feedback = history.history['val_' + metric_name][-1]
+        return model, feedback
+
+
+class ImageSupervised(auto_pipeline.AutoPipeline):
```
fchollet: This is a base class, right? Perhaps `SupervisedImagePipeline` would be a more explicit name.

autokeras/auto/processor.py
```diff
@@ -0,0 +1,81 @@
+import numpy as np
+
+
+class OneHotEncoder:
```
fchollet: `(object)`

autokeras/auto/processor.py
```diff
@@ -0,0 +1,81 @@
...
+    """A class that can format data.
+
+    This class provides ways to transform data's classification label into
+    vector.
+
+    Attributes:
```
fchollet: Use markdown conventions:

```
# Attributes
```

autokeras/auto/processor.py
```diff
@@ -0,0 +1,81 @@
...
+          n_classes: The number of classes in the classification problem.
+          labels: The number of labels.
+          label_to_vec: Mapping from label to vector.
+          int_to_label: Mapping from int to label.
+    """
+
+    def __init__(self):
+        """Initialize a OneHotEncoder"""
+        self.data = None
+        self.n_classes = 0
```
fchollet: `num_`

autokeras/auto/processor.py
fchollet: Keep in mind that this whole file will end up deprecated with preprocessing layers and tf.data. Don't put too much work into it.

autokeras/auto/processor.py
```diff
@@ -0,0 +1,81 @@
...
+    """
+
+    def __init__(self):
+        self.max_val = None
+        self.mean = None
+        self.std = None
+
+    def fit(self, data):
+        self.max_val = data.max()
+        data = data / self.max_val
```
fchollet: Prefer following the API of the Normalization layer (mean and stddev normalization) to avoid issues when converting to a preprocessing layer.

autokeras/auto/image.py
jhfjhfj1: This is already configurable from fit function of AutoModel. fit_kwargs is the **kwargs of fit function of AutoModel.

autokeras/auto/image.py
jhfjhfj1: I will submit another pull request for changing this one.

jhfjhfj1:
The comments are all addressed except for using tf.data. It is ready for another review. Thanks. @fchollet

autokeras/auto/auto_model.py
```diff
@@ -96,21 +84,29 @@ def __init__(self,
         self._hypermodels = []
         self._hypermodel_to_id = {}
         self._build_network()
+        self.tuner = tuner.SequentialRandomSearch(
+            self,
+            objective=self._get_metrics())
 
-    def build(self, hp):
+    def build(self, hp, **kwargs):
```
fchollet: Where are `**kwargs` used?

If another piece of code need to programatically call `build`, how can it know what arguments to pass?

autokeras/auto/auto_model.py
```diff
@@ -146,20 +143,24 @@ def _build_network(self):
...
+                # Check at least one output node of the hypermodel is in the
+                # interested nodes.
+                if not any([output_node in self._node_to_id for output_node in
+                            hypermodel.outputs]):
                     continue
                 self._add_hypermodel(hypermodel)
                 for output_node in hypermodel.outputs:
                     # The node is not visited and in interested nodes.
-                    if output_node not in visited_nodes and output_node in self._node_to_id:
+                    if output_node not in visited_nodes \
```
fchollet: Prefer

```
if (output_node not in visited_nodes and
     output_node in self._node_to_id):
```

autokeras/auto/auto_model.py
jhfjhfj1: This is removed since we are not using the **kwargs in the build function.

PR #655, opened 2019-06-09
[MRG] Improve DenseBlock

Davidsirui:
Add some other kind of layers to DenseBlock

### DIFF
--- a/autokeras/hypermodel/hyper_block.py
+++ b/autokeras/hypermodel/hyper_block.py
@@ -41,16 +41,9 @@ class DenseBlock(HyperBlock):
         output_node = Flatten().build(hp, output_node)
 
         for i in range(hp.Choice('num_layers', [1, 2, 3], default=2)):
-            if hp.Choice('layer_category', ['Dense','BN','Dropout','Activation'], default='Dense') == 'Dense':
                 output_node = tf.keras.layers.Dense(hp.Choice('units_{i}'.format(i=i),
                                                               [16, 32, 64],
                                                               default=32))(output_node)
-            elif hp.Choice('layer_category', ['Dense','BN','Dropout','Activation'], default='Dense') == 'BN':
-                output_node = tf.keras.layers.BatchNormalization()(output_node)
-            elif hp.Choice('layer_category', ['Dense','BN','Dropout','ACtivation'], default='Dense') == 'Dropout':
-                output_node = tf.keras.layers.Dropout(rate=hp.Choice('dropout_rate',[0.1, 0.3, 0.5],default=0.5))(output_node)
-            else:
-                output_node = tf.keras.layers.Activation(hp.Choice('activate_category', ['softmax','relu','tanh','sigmoid'], default='relu'))(output_node)
         return output_node
 
 
--- a/autokeras/hypermodel/hyper_block.py
+++ b/autokeras/hypermodel/hyper_block.py
@@ -39,22 +39,29 @@ class DenseBlock(HyperBlock):
         input_node = layer_utils.format_inputs(inputs, self.name, num=1)[0]
         output_node = input_node
         output_node = Flatten().build(hp, output_node)
-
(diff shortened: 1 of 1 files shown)


autokeras/hypermodel/hyper_block.py
```diff
@@ -41,9 +41,20 @@ def build(self, hp, inputs=None):
         output_node = Flatten().build(hp, output_node)
 
         for i in range(hp.Choice('num_layers', [1, 2, 3], default=2)):
-            output_node = tf.keras.layers.Dense(hp.Choice('units_{i}'.format(i=i),
-                                                          [16, 32, 64],
-                                                          default=32))(output_node)
+            if hp.Choice('layer_category', ['Dense','BN','Dropout','Activation'], default='Dense') == 'Dense':
```
jhfjhfj1: As long as you use the same name of 'layer_category', the value of it will always be the same. So it would be all dense, or all dropout, or all activation.
I think we can do it like the following:

`for ... 
   layer_stack = hp.Choice('layer_stack', ['dense-bn-act', 'dense-act', 'act-bn-dense']
   if layer_stack == ...:
   elif layer_stack == ...:

autokeras/hypermodel/hyper_block.py
```diff
@@ -41,9 +41,20 @@ def build(self, hp, inputs=None):
...
+                                                              default=32))(output_node)
+            elif hp.Choice('layer_category', ['Dense','BN','Dropout','Activation'], default='Dense') == 'BN':
+                # TODO: Insert BN layer
+                output_node = tf.keras.layers.BatchNormalization()(output_node)
+            elif hp.Choice('layer_category', ['Dense','BN','Dropout','ACtivation'], default='Dense') == 'Dropout':
+                # TODO: Insert Dropout layer
+                output_node = tf.keras.layers.Dropout(rate=hp.Choice('dropout_rate',[0.1, 0.3, 0.5],default=0.5))(output_node)
+            else:
+                # TODO: Insert Activation layer
+                output_node = tf.keras.layers.Activation(hp.Choice('activate_category', ['softmax','relu','tanh','sigmoid'], default='relu'))(output_node)
```
jhfjhfj1: This is the correct behavior.
Currently, we want all the activation layers of the same type.
To make the code less ambiguous, please extract "activation_category" as a variable and put it before the loop.
Use the variable in the loop instead of directly calling hp....

autokeras/hypermodel/hyper_block.py
```diff
@@ -41,9 +41,20 @@ def build(self, hp, inputs=None):
...
+                                                              [16, 32, 64],
+                                                              default=32))(output_node)
+            elif hp.Choice('layer_category', ['Dense','BN','Dropout','Activation'], default='Dense') == 'BN':
+                # TODO: Insert BN layer
+                output_node = tf.keras.layers.BatchNormalization()(output_node)
+            elif hp.Choice('layer_category', ['Dense','BN','Dropout','ACtivation'], default='Dense') == 'Dropout':
+                # TODO: Insert Dropout layer
+                output_node = tf.keras.layers.Dropout(rate=hp.Choice('dropout_rate',[0.1, 0.3, 0.5],default=0.5))(output_node)
+            else:
+                # TODO: Insert Activation layer
```
jhfjhfj1: TODO means something to be done. I think it is already done in the code. So we don't need to use all the TODOs in here

autokeras/hypermodel/hyper_block.py
Davidsirui: If we use this kind of architecture, it means the search space is much less than before. Because we are just stacking these 'daense-bn-act'. So  I think we just give them a different name is OK. For example, 'layer_category_Dense','layer_category_BN'. So this will totally make a different. Each loop, we will get a different layer.

jhfjhfj1: You are right that the search space is restricted. However, that is what we want for now. Currently, the search algorithms are not efficient enough. So we want to restrict the space to a small one that most of the architectures in it are good.

Issue #653, opened 2019-06-07
XceptionBlock


jhfjhfj1:

jhfjhfj1:
@eustomaqua Here is an example released yesterday, you can change the the search space according to it. Thanks. https://github.com/keras-team/keras-tuner/tree/master/kerastuner/applications/tunable_xception

PR #656, opened 2019-06-10
[MRG] XceptionBlock implemented (#653)

eustomaqua:
XceptionBlock #653 

### Pull Request Status 
ready to be merged

The Xception Block will return a similar architecture as original Xception except without the last (optional) fully connected layer(s) and logistic regression, with a half, an identical, or a double size of the original one.

### DIFF
--- a/autokeras/hypermodel/hyper_block.py
+++ b/autokeras/hypermodel/hyper_block.py
@@ -101,16 +101,32 @@ class Merge(HyperBlock):
 
 
 class XceptionBlock(HyperBlock):
+    """ XceptionBlock
+    An Xception structure, used for specifying your model with specific datasets.
+
+    The original Xception architecture is from https://arxiv.org/abs/1610.02357.
+    The data first goes through the entry flow, then through the middle flow which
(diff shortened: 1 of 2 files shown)


autokeras/hypermodel/hyper_block.py
```diff
@@ -101,9 +101,111 @@ def build(self, hp, inputs=None):
...
+        input_node = layer_utils.format_inputs(inputs, self.name, num=1)[0]
+        output_node = input_node
+        channel_axis = 1 if tf.keras.backend.image_data_format() == 'channels_first' else -1
+
+        # To decide the size of the architecture
+        filter_entry = hp.Choice('num_filters', [16, 32, 64], default=32)
+        filter_middle = int(filter_entry * 22.75)
+        filter_exit = int(filter_entry * 48)
+        # To decide the pooling layer in the end
+        last_pooling = hp.Choice('pooling', ['avg', 'max'], default='avg')
```
jhfjhfj1: We don't need the pooling in the block.
So just delete the pooling layers and this variable.

PR #658, opened 2019-06-10
[MRG] sequential auto model implemented

jhfjhfj1:

### DIFF
--- a/autokeras/__init__.py
+++ b/autokeras/__init__.py
@@ -1,2 +1,2 @@
 from autokeras.hypermodel import *
-from autokeras.auto.auto_model import AutoModel
+from autokeras.auto import *
--- a/autokeras/auto/auto_model.py
+++ b/autokeras/auto/auto_model.py
@@ -8,6 +8,7 @@ from autokeras.hypermodel import hypermodel
 from autokeras.hypermodel import hyper_head
 from autokeras.hypermodel import hyper_node
 from autokeras import layer_utils
+from autokeras import tuner
 
 
 class AutoModel(hypermodel.HyperModel):
@@ -33,6 +34,7 @@ class AutoModel(hypermodel.HyperModel):
         self._nodes = []
         self._built = False
         self.tuner = None
+        self.tuner = tuner.SequentialRandomSearch(self, objective=self._get_metrics())
 
     def build(self, hp, **kwargs):
         raise NotImplementedError
@@ -136,7 +138,6 @@ class GraphAutoModel(AutoModel):
 
         model = tf.keras.Model([real_nodes[self._node_to_id[input_node]] for input_node in self.inputs],
                                [real_nodes[self._node_to_id[output_node]] for output_node in self.outputs])
-
 
         return self._compiled(hp, model)
 
--- a/autokeras/auto/auto_model.py
+++ b/autokeras/auto/auto_model.py
@@ -4,7 +4,9 @@ import numpy as np
 import tensorflow as tf
 from tensorflow.python.util import nest
 
-from autokeras.hypermodel import hypermodel, hyper_head, hyper_node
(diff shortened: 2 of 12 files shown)


autokeras/auto/auto_model.py
```diff
@@ -81,6 +91,18 @@ def _get_metrics(self):
             metrics += metrics_list
         return metrics
 
+    def _compiled(self, hp, model):
+        optimizer = hp.Choice('optimizer',
```
omalleyt12: Why are these choices hard-coded?

autokeras/auto/auto_model.py
```diff
@@ -28,17 +28,27 @@ def __init__(self, **kwargs):
...
     def fit(self,
             x=None,
             y=None,
             validation_data=None,
             trials=None,
             **kwargs):
+
+        if not self._built:
+            self._hyper_build()
+
```
omalleyt12: Should we set `self._built` here? Can the use override `self._hyper_build()` / `self.build()` and maybe not call `self.built = True`?

autokeras/auto/auto_model.py
```diff
@@ -4,7 +4,7 @@
 import tensorflow as tf
 from tensorflow.python.util import nest
 
-from autokeras.hypermodel import hypermodel, hyper_head
+from autokeras.hypermodel import hypermodel, hyper_head, hyper_node
```
omalleyt12: Prefer one import per line

tests/test_auto_model.py
```diff
@@ -123,3 +123,15 @@ def test_auto_model_basic():
...
 
     assert result.shape == (100, 1)
+
+
+def test_sequential_auto_model_init():
+    x_train = np.random.rand(100, 32)
+    y_train = np.random.rand(100)
+
+    auto_model = SequentialAutoModel([DenseBlock(), RegressionHead()])
+    const.Constant.NUM_TRAILS = 2
```
omalleyt12: Typo: TRIALS?

omalleyt12: Is TRIALS a global constant? Maybe it could be an argument passed to `fit` instead?

autokeras/auto/auto_model.py
jhfjhfj1: Do you mean the search space is small? I think this will be changed the later to be more flexible. Currently, it is just a selection among these optimizers.

autokeras/auto/auto_model.py
jhfjhfj1: I am not sure how to change it. We can discuss during our meeting tomorrow.

jhfjhfj1:
This one is also ready for review. Thanks. @fchollet

autokeras/__init__.py
```diff
@@ -1,2 +1,2 @@
 from autokeras.hypermodel import *
-from autokeras.auto.auto_model import AutoModel
+from autokeras.auto import *
```
fchollet: Prefer importing symbols explicitly, one at a time

autokeras/auto/auto_model.py
```diff
@@ -28,30 +30,44 @@ def __init__(self, **kwargs):
         super().__init__(**kwargs)
         self.inputs = []
         self.outputs = []
+        self._hypermodels = []
+        self._nodes = []
+        self._built = False
         self.tuner = None
+        self.tuner = tuner.SequentialRandomSearch(self, objective=self._get_metrics())
```
fchollet: You should now be able to add Keras Tuner (installed from GitHub) as a dependency of AutoKeras.

In this case, since this is a base class, you shouldn't need to set a concrete tuner.

jhfjhfj1:
Since the master has changed too much and the new API is not the same as the current sequential, I will close this pull request and send a new one.

PR #659, opened 2019-06-11
Improve DenseBlock

Davidsirui:


jhfjhfj1:
@Davidsirui we should continue on the original pull request instead of creating a new one. Thanks. Remember to use local IDEs for the development which would be more efficient.

PR #663, opened 2019-06-12
[MRG] Text Classifier and Regressor

### LINKED ISSUE #654: TextClassifier and TextRegressor
### Feature Description

Follow the new version of imagesupervised to do the work.
You can create a simple TextBlock() to do the job and we can change it later.

satyakesav:
This pull request resolves #654. 

### Pull Request Status 
Ready to be merged

Implement Text Classifier and Regressor classes.


satyakesav:
> Hi Satya,
> The code generally looks OK. There some more things to be added.
> 
> 1. We would like the input to be raw text data (strings). So you may add embedding layer in TextBlock and connect the output of embedding to RNNblock.
> 2. You may convert the text input to numerical vectors in text tuner just like what we did for image data augmentation.
> 3. For the current version, is there any difference between the TextTuner's _run and the original one?
> 
> Thanks.

Hi Haifeng,

I initially planned to incorporate point 1 in future PRs. But, I can do it in this PR. So, I will address both points 1 and 2 in the current PR. Regarding the third point, there is a slight difference between TextTuner's _run and the original one. In the original, model.compile() is not called without the arguments loss/optimizer/metrics and when I tested this, it threw errors. So, I planned to include a separate _run function for now in this PR. From the official documentation, I see that optimizer is a mandatory argument but it is not the case in the original method. So, Is there something wrong with the code or am I missing on something?

jhfjhfj1:
There are major changes in the master branch. Please change the pull request accordingly. If too many conflicts, you can close the pull request, start a new branch from master, copy and paste your code to it, and submit a new one. Thanks.

PR #667, opened 2019-06-13
[MRG] Fix review comments and enforce 3-D inputs

PraveenVenugopal:

Fix review comments and allow only 3-D time series inputs from the user for all RNN blocks


autokeras/layer_utils.py
```diff
@@ -26,6 +26,10 @@ def format_inputs(inputs, name=None, num=None):
     return inputs
 
 
+def get_block(choice):
```
jhfjhfj1: rename to get_rnn_block

autokeras/hypermodel/hyper_block.py
```diff
@@ -69,34 +70,22 @@ class RNNBlock(HyperBlock):
 
     def build(self, hp, inputs=None):
         input_node = layer_utils.format_inputs(inputs, self.name, num=1)[0]
+        feature_size = input_node.shape[2]
```
jhfjhfj1: add a check first, if len(input_node.shape) < 3: raise ValueError('The input to RNNBlock must have at least 3 dimensions').

tests/test_hyper_block.py
```diff
@@ -3,8 +3,8 @@
 from autokeras.auto.auto_model import *
 from autokeras import const
 
-from autokeras.hypermodel.hyper_block import XceptionBlock, RNNBlock
-from autokeras.hypermodel.hyper_head import ClassificationHead, RegressionHead
+from autokeras.hypermodel.hyper_block import *
+from autokeras.hypermodel.hyper_head import *
```
jhfjhfj1: This is test_hyper_block so we only do hyper_block import *, but not import * for other modules.

autokeras/hypermodel/hyper_block.py
PraveenVenugopal: Should we check for < 3 or strictly == 3. To enforce 3-D shape i believe we need the second

autokeras/layer_utils.py
PraveenVenugopal: Sure.

tests/test_hyper_block.py
PraveenVenugopal: Okay. got it

PR #684, opened 2019-06-27
[MRG] Automodel interface update

jhfjhfj1:


autokeras/auto/auto_model.py
```diff
@@ -214,9 +152,107 @@ def _add_node(self, input_node):
...
+                               'sgd'])
+
+        model.compile(optimizer=optimizer,
+                      metrics=self._get_metrics(),
+                      loss=self._get_loss())
+
+        return model
+
+
+class AutoModel(GraphAutoModel):
```
tonyabracadabra: At least from the naming perspective, `GraphAutoModel`  should inherit `AutoModel`.

autokeras/auto/auto_model.py
```diff
@@ -214,9 +152,107 @@ def _add_node(self, input_node):
...
+                 outputs,
+                 **kwargs):
+        inputs = layer_utils.format_inputs(inputs)
+        outputs = layer_utils.format_inputs(outputs)
+        middle_nodes = [input_node.related_block()(input_node)
+                        for input_node in inputs]
+        if len(middle_nodes) > 1:
+            output_node = hyper_block.Merge()(middle_nodes)
+        else:
+            output_node = middle_nodes[0]
```
tonyabracadabra: Why do we need this? Can't you let `hyper_block.Merge()` taking care of the merge for a list that contains one node?

autokeras/hypermodel/hyper_node.py
```diff
@@ -26,20 +27,34 @@ def __init__(self, **kwargs):
...
-class ImageInput(Node):
 
-    def __init__(self, **kwargs):
-        super().__init__(**kwargs)
+class ImageInput(Input):
 
-    def build(self, hp):
-        return tf.keras.Input(shape=self.shape)
+    @staticmethod
+    def related_block():
```
tonyabracadabra: Maybe you want to make `related_block` an abstract static method under `Input`, then extract out an `GeneralInput` inherits under `Input`

autokeras/auto/auto_model.py
```diff
@@ -214,9 +152,107 @@ def _add_node(self, input_node):
...
+    def predict(self, x, **kwargs):
+        """Predict the output for a given testing data. """
+        return self.tuner.get_best_models(1)[0].predict(x, **kwargs)
+
+    def _get_metrics(self):
+        metrics = []
+        for metrics_list in [output_node.in_hypermodels[0].metrics for
+                             output_node in self.outputs
+                             if isinstance(output_node.in_hypermodels[0],
+                                           hyper_head.HyperHead)]:
```
tonyabracadabra: I don't think there is a need for two for loops.

autokeras/auto/auto_model.py
```diff
@@ -214,9 +152,107 @@ def _add_node(self, input_node):
...
+
+        # TODO: allow early stop if epochs is not specified.
+        self.tuner.search(x=x,
+                          y=y,
+                          validation_data=validation_data,
+                          **kwargs)
+
+    def predict(self, x, **kwargs):
+        """Predict the output for a given testing data. """
+        return self.tuner.get_best_models(1)[0].predict(x, **kwargs)
```
tonyabracadabra: I've always been cautious in indexing without prechecking the boundary, maybe it is taken cared already in keras tuner.

autokeras/auto/auto_model.py
```diff
@@ -214,9 +153,107 @@ def _add_node(self, input_node):
...
+                y_input = np.reshape(y_input, y_input.shape + (1,))
+            output_node.shape = y_input.shape[1:]
+            output_node.in_hypermodels[0].output_shape = output_node.shape
+
+        # Prepare the dataset
+        if validation_data is None:
+            (x, y), (x_val, y_val) = layer_utils.split_train_to_valid(x, y)
+            validation_data = x_val, y_val
+
+        self.tuner = kerastuner.RandomSearch(
```
fchollet: All of these arguments are known in the constructor; as such it should be possible to construct the tuner in the constructor rather than in `fit`, no?

autokeras/auto/auto_model.py
```diff
@@ -214,9 +153,107 @@ def _add_node(self, input_node):
...
+        for x_input, input_node in zip(x, self.inputs):
+            input_node.shape = x_input.shape[1:]
+        for y_input, output_node in zip(y, self.outputs):
+            if len(y_input.shape) == 1:
+                y_input = np.reshape(y_input, y_input.shape + (1,))
+            output_node.shape = y_input.shape[1:]
+            output_node.in_hypermodels[0].output_shape = output_node.shape
+
+        # Prepare the dataset
+        if validation_data is None:
```
fchollet: `validation_split` and `validation_data` are already valid arguments in `model.fit`. Besides, handling validation data correctly is very complicated (here you assume numpy arrays, but this is only one of the possible cases). It is much simpler to just pass the fit_args and fit_kwargs blindly to `model.fit()` without any special handling of the validation data.

autokeras/auto/auto_model.py
fchollet: I suspect we should have a a BaseAutoModel (implementer-facing) to handle abstract logic that does not rely on the graph API, and an AutoModel subclass that is public-facing and implements the graph API.

autokeras/auto/auto_model.py
```diff
@@ -4,96 +4,45 @@
...
-        inputs: A HyperModel instance. The input node of a the AutoModel.
-        outputs: A HyperModel instance. The output node of the AutoModel.
-        hypermodel: An instance of HyperModelWrap connecting from the inputs to the
-            outputs.
-        tuner: An instance of Tuner.
+        inputs: A list of or a HyperNode instances.
+            The input node(s) of the GraphAutoModel.
+        outputs: A list of or a HyperNode instances.
+            The output node(s) of the GraphAutoModel.
+        trials: Int. The number of different models to try.
```
fchollet: Should this be renamed to `max_trials` across the board? It may not be an exact number, but rather an upper bound, because the tuner or oracle may interrupt the tuning process (e.g. if the search space is too small).

autokeras/auto/auto_model.py
jhfjhfj1: In this PR, it can, but with the preprocessing blocks it is not. The tuner will do one round of build during initialization, which will need to call the preprocessing blocks to transform the data, which means we have to see the training data.

autokeras/auto/auto_model.py
jhfjhfj1: If we don't handle it, does it mean the users have to provide the validation data? If so we would need to handle it even it is complicated.

jhfjhfj1:
For the AutoModel and GraphAutoModel, for now, the logic is really AutoModel is a subclass of GraphAutoModel, since it needs everything in the GraphAutoModel. We can look back to it when we have the metamodel ready, to see if there are common logic to extract to the base class. @fchollet

jhfjhfj1:
I have added the issues in https://github.com/keras-team/autokeras/projects/3#card-23314326 about 1. the AutoModelBase (#689) and when to initialize the tuner (#688) and whether we should handle validation data in AutoKeras (#690).
I will merge the pull request just not to block the rest of the development, we can continue discussion on these in the issues.

PR #686, opened 2019-06-28
Detail Changes for Blocks

jhfjhfj1:


autokeras/hypermodel/hyper_block.py
```diff
@@ -41,51 +43,43 @@ def build(self, hp, inputs=None):
...
+                output_node = tf.keras.layers.ReLU()(output_node)
+                output_node = tf.keras.layers.Dropout(dropout_rate)(output_node)
         return output_node
 
 
 class RNNBlock(HyperBlock):
+    """ An RNN HyperBlock.
+
+    Attributes:
+        bidirectional: Boolean. If not provided, it would be a tunable variable.
```
fchollet: Why make it configurable, i.e. in what workflows do you expect users to want to change this via a constructor argument (as opposed to, say, the use of BN in DenseBlock)?

autokeras/hypermodel/hyper_block.py
```diff
@@ -321,38 +366,70 @@ def func(x):
 
 class Flatten(HyperBlock):
 
+    def build(self, hp, inputs=None):
+        input_node = layer_utils.format_inputs(inputs, self.name, num=1)[0]
```
fchollet: Is there any difference between this and `tf.nest.flatten()`? The above is not very readable on its own (i.e. a reader of the code wouldn't immediately know what this does).

autokeras/hypermodel/hyper_block.py
```diff
@@ -321,38 +366,70 @@ def func(x):
...
+        # No need to reduce.
+        if len(output_node.shape) <= 2:
+            return output_node
+
+        reduction_type = hp.Choice('reduction_type',
+                                   ['flatten',
+                                    'max',
+                                    'ave',
+                                    'min'],
+                                   default='global_ave')
```
fchollet: The default is not among the choices.

autokeras/hypermodel/hyper_block.py
```diff
@@ -321,38 +366,70 @@ def func(x):
...
 
+        # No need to reduce.
+        if len(output_node.shape) <= 2:
+            return output_node
+
+        reduction_type = hp.Choice('reduction_type',
+                                   ['flatten',
+                                    'max',
+                                    'ave',
+                                    'min'],
```
fchollet: Perhaps these choices should be perfixed with `global_` to be consistent with the values in `SpatialReduction`.

autokeras/hypermodel/hyper_head.py
```diff
@@ -17,26 +17,28 @@ def build(self, hp, inputs=None):
 
 class ClassificationHead(HyperHead):
 
-    def __init__(self, **kwargs):
+    def __init__(self, binary=False, **kwargs):
```
fchollet: Shouldn't the number of classes be configurable here?

autokeras/hypermodel/hyper_head.py
```diff
@@ -17,26 +17,28 @@ def build(self, hp, inputs=None):
...
         super().__init__(**kwargs)
+        self.binary = binary
         if not self.metrics:
             self.metrics = ['accuracy']
         if not self.loss:
-            self.loss = 'categorical_crossentropy'
+            if binary:
+                self.loss = 'binary_crossentropy'
+                self.output_shape = (1,)
```
fchollet: Binary classification is valid for outputs of various sizes, not just scalar outputs

autokeras/hypermodel/hyper_head.py
```diff
@@ -49,11 +51,9 @@ def __init__(self, **kwargs):
             self.loss = 'mean_squared_error'
 
     def build(self, hp, inputs=None):
+        print(inputs)
```
fchollet: Debug statement?

autokeras/hypermodel/hyper_block.py
```diff
@@ -102,40 +96,91 @@ def build(self, hp, inputs=None):
...
-            return_sequences = False if i == choice_of_layers - 1 else True
-            bidirectional_block = tf.keras.layers.Bidirectional(
-                in_layer(feature_size, activation='tanh',
-                         return_sequences=return_sequences))
-            output_node = bidirectional_block(output_node)
+        bidirectional = self.bidirectional
+        if bidirectional is None:
+            bidirectional = hp.Choice('bidirectional',
+                                      [True, False],
+                                      default=True)
```
tonyabracadabra: This line is potentially making `bidirectional` and `self.bidirectional` inconsistent

autokeras/hypermodel/hyper_block.py
```diff
@@ -102,40 +96,91 @@ def build(self, hp, inputs=None):
...
 
-        output_node = Flatten().build(hp, output_node)
+        return_sequences = self.return_sequences
+        if return_sequences is None:
+            return_sequences = hp.Choice('return_sequences',
+                                         [True, False],
+                                         default=True)
```
tonyabracadabra: same as above, I suggest removing `return_sequence` and `bidirectional` variable, and update `self.return_sequence` and `self.bidirectional` when they are `None`

autokeras/hypermodel/hyper_block.py
```diff
@@ -102,40 +96,91 @@ def build(self, hp, inputs=None):
...
+        block_type = hp.Choice('block_type',
+                               ['resnet', 'xception', 'vanilla'],
+                               default='resnet')
+
+        if block_type == 'resnet':
+            output_node = ResNetBlock().build(hp, output_node)
+        elif block_type == 'xception':
+            output_node = XceptionBlock().build(hp, output_node)
+        elif block_type == 'vanilla':
+            output_node = ConvBlock().build(hp, output_node)
```
tonyabracadabra: I prefer having a `string` to `Block` map `block_map` then just do `block_map[block_type].build(hp, output_node)`

autokeras/hypermodel/hyper_block.py
```diff
@@ -41,51 +43,43 @@ def build(self, hp, inputs=None):
...
+                output_node = tf.keras.layers.Dropout(dropout_rate)(output_node)
         return output_node
 
 
 class RNNBlock(HyperBlock):
+    """ An RNN HyperBlock.
+
+    Attributes:
+        bidirectional: Boolean. If not provided, it would be a tunable variable.
+        return_sequences: Boolean.  If not provided, it would be a tunable variable.
```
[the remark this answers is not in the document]
jhfjhfj1: If it is text classification, it can either return a sequence and reduce the dimension, or directly return the vector. I think they can be alternatives.

autokeras/hypermodel/hyper_head.py
jhfjhfj1: If it is configurable, it would only be a security check for the targets user passed contains the same number of classes.

autokeras/hypermodel/hyper_head.py
jhfjhfj1: This part I don't really understand. We will discuss during the meeting.

autokeras/hypermodel/hyper_block.py
jhfjhfj1: This part I am not sure. We may discuss during the meeting.

autokeras/meta_model.py
```diff
@@ -21,13 +21,13 @@ def assemble(inputs, outputs, dataset):
...
             assemblers.append(TextAssembler())
-        if isinstance(input_node, hyper_node.ImageInput):
+        if isinstance(input_node, node.ImageInput):
             assemblers.append(ImageAssembler())
-        if isinstance(input_node, hyper_node.StructuredInput):
+        if isinstance(input_node, node.StructuredInput):
             assemblers.append(StructuredDataAssembler())
-        if isinstance(input_node, hyper_node.TimeSeriesInput):
+        if isinstance(input_node, node.TimeSeriesInput):
             assemblers.append(TimeSeriesAssembler())
```
tonyabracadabra: Why don't you put a class variable under `Input`? then you can do
```
assemblers.append(input_node.assembler())
```
it is not only saving you few lines of codes but also easier to manage the dependencies between `Input` and `Assembler`

autokeras/hypermodel/block.py
```diff
@@ -0,0 +1,595 @@
...
+        pass
+
+
+class TimeSeriesBlock(HyperBlock):
+
+    def build(self, hp, inputs=None):
+        pass
+
+
+class GeneralBlock(HyperBlock):
```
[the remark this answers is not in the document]
jhfjhfj1: The user may not specify the type of the input since their input may not fall into any of the categories we defined. They would want this general block if they don't know anything about what architecture to use.

autokeras/auto_model.py
```diff
@@ -24,27 +24,31 @@ class AutoModel(kerastuner.HyperModel):
...
 
     def __init__(self,
                  inputs,
                  outputs,
-                 max_trials=None,
-                 directory=None,
-                 **kwargs):
-        super().__init__(**kwargs)
+                 max_trials=100,
+                 directory='./tmp',
```
fchollet: In Keras Tuner it defaults to the current directory (where a new subfolder for the current project gets created).

autokeras/auto_model.py
```diff
@@ -24,27 +24,31 @@ class AutoModel(kerastuner.HyperModel):
...
+                 seed=None):
+        super().__init__(name='AutoModel')
```
fchollet: By convention, names of things of things are usually 1) configurable, 2) snake case, like `auto_model`.

autokeras/auto_model.py
```diff
@@ -289,7 +295,13 @@ def prepare_data(self, x, y, validation_data, validation_split):
         return dataset, validation_data
 
     def predict(self, x, batch_size=32, **kwargs):
-        """Predict the output for a given testing data. """
+        """Predict the output for a given testing data.
+
+        # Arguments:
```
fchollet: No `:`

autokeras/hypermodel/block.py
```diff
@@ -0,0 +1,599 @@
...
+                                                        [True, False],
+                                                        default=True)
+        layer_type = self.layer_type or hp.Choice('layer_type',
+                                                  ['gru', 'lstm'],
+                                                  default='lstm')
+        num_layers = self.num_layers or hp.Choice('num_layers',
+                                                  [1, 2, 3],
+                                                  default=2)
+        rnn_layers = {
+            'vanilla': tf.keras.layers.SimpleRNN,
```
fchollet: This is not included in the `layer_type` hp, and does not seem to be useful in general.

autokeras/hypermodel/block.py
```diff
@@ -0,0 +1,599 @@
...
+        initial_strides: Int. If left unspecified, it will be tuned automatically.
+        num_residual_blocks: Int. If left unspecified, it will be tuned
+            automatically.
+        pooling: Str. 'ave', 'flatten', or 'max'. If left unspecified, it will be
+            tuned automatically.
+    """
+
+    def __init__(self,
+                 activation=None,
+                 conv2d_num_filters=None,
```
fchollet: Why not just `num_filters`?

autokeras/hypermodel/block.py
```diff
@@ -0,0 +1,599 @@
...
+
+    This XceptionBlock returns a similar architecture as Xception except without
+    the last (optional) fully connected layer(s) and logistic regression.
+    The size of this architecture could be decided by `HyperParameters`, to get an
+    architecture with a half, an identical, or a double size of the original one.
+
+    # Arguments
+        activation: Str. 'selu' or 'relu'. If left unspecified, it will be tuned
+            automatically.
+        conv2d_num_filters: Int. If left unspecified, it will be tuned automatically.
```
fchollet: Not describing an argument in detail is ok if the name of the argument is explicit on its own. In this case, though, it is not clear from the name only what this argument does.

autokeras/hypermodel/block.py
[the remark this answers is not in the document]
fchollet: How would that be supported in practice?

autokeras/hypermodel/block.py
```diff
@@ -0,0 +1,599 @@
...
+
+
+class EmbeddingBlock(HyperBlock):
+    """Word embedding block for sequences.
+
+    The input should be tokenized sequences with the same length, where each element
+    of a sequence should be the index of the word.
+
+    # Arguments
+        pretrained: Str. 'no pretrained', 'glove', 'fasttext' or 'word2vec'. Use
```
fchollet: `pretrained='no pretrained'` seems strange; let's go with `pretraining={'none'|'glove'|etc}`

autokeras/auto_model.py
```diff
@@ -471,18 +489,35 @@ class GraphAutoModel(AutoModel):
...
+            AutoModel in the current directory.
+        seed: Int. Random seed.
     """
 
-    def __init__(self, *args, **kwargs):
-        super(GraphAutoModel, self).__init__(*args, **kwargs)
+    def __init__(self,
+                 inputs,
+                 outputs,
+                 name='auto_graph_model',
```
fchollet: Is this supposed to be 'graph_auto_model'?

autokeras/auto_model.py
```diff
@@ -471,18 +489,35 @@ class GraphAutoModel(AutoModel):
...
+                 max_trials=100,
+                 directory='./tmp',
```
fchollet: Update this

PR #691, opened 2019-07-01
MetaModel for text data

jhfjhfj1:
resolves #000


autokeras/hypermodel/hyper_block.py
```diff
@@ -156,29 +156,46 @@ def build(self, hp, inputs=None):
 
 class ConvBlock(HyperBlock):
 
+    def __init__(self, separable=None, **kwargs):
```
fchollet: I am in general confused as to why some arguments are present in the constructor and as hps, while others are only hps.

We should follow a consistent convention here.

How about: all arguments are listed in the constructor, defaulting to None, and if they're None they get selected via a hp?

autokeras/meta_model.py
```diff
@@ -0,0 +1,132 @@
...
+    # for text data we just follow the rules on that page.
+    # for image, use the num_instance to determine the range of the sizes of the
+    # resnet and xception
+    # use the image size to determine how the down sampling works, e.g. pooling.
+
+
+class Assembler(object):
+    """Assembles the HyperModels together in AutoModel for different inputs."""
+
+    def assemble(self, x, input_node, y, outputs):
```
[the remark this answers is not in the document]
fchollet: Also kinda confusing why the signature is different from that of `assemble(inputs, outputs, x, y)`

autokeras/hypermodel/hyper_block.py
jhfjhfj1: Good idea. Since it requires massive changes to all the hyperblocks, it would be a separate pull request.

autokeras/meta_model.py
```diff
@@ -0,0 +1,132 @@
...
+    return nest.flatten([output_blocks(output_node)
+                         for output_blocks in outputs])
+    # all inputs, all train_x y, all heads
+    # for text data we just follow the rules on that page.
+    # for image, use the num_instance to determine the range of the sizes of the
+    # resnet and xception
+    # use the image size to determine how the down sampling works, e.g. pooling.
+
+
+class Assembler(object):
```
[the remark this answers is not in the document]
jhfjhfj1: After the recent change of using `update` function, it needs a stateful thing to record the statistics. So we would keep this OO design

autokeras/meta_model.py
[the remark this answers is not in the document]
jhfjhfj1: Now the update and assemble are two different functions, so we can avoid this problem.

autokeras/hypermodel/processor.py
```diff
@@ -107,7 +167,91 @@ def transform(self, hp, data):
...
+        return (x - self.mean) / self.std
+
+    def output_types(self):
+        return tf.float64,
+
+    def output_shape(self):
+        return self.mean.shape
+
+
+class TextToSequenceVector(HyperPreprocessor):
```
fchollet: It isn't very clear what a "sequence vector" would be. Perhaps `TextToIntSequence` would be clearer.

autokeras/hypermodel/processor.py
```diff
@@ -1,33 +1,94 @@
...
+
+        The output shape is needed to build the Keras Model from the AutoModel.
+        The output shape of the preprocessor is the input shape of the Keras Model.
+
+        Returns:
+            A tuple of ints or a TensorShape.
+        """
+        raise NotImplementedError
+
+    def post_fit(self):
```
fchollet: Maybe `finalize` would be clearer. There is not `fit` method in this class. `finalize` and "finalizer" are fairly standard concepts. https://en.wikipedia.org/wiki/Finalizer

tests/test_task.py
```diff
@@ -0,0 +1,67 @@
...
+    x_train = np.array(x_train, dtype=np.str)
+    x_test = np.array(x_test, dtype=np.str)
+    return (x_train, y_train), (x_test, y_test)
+
+
+def test_text_classifier(tmp_dir):
+    (train_x, train_y), (test_x, test_y) = imdb_raw()
+    clf = ak.TextClassifier(directory=tmp_dir, max_trials=2)
+    clf.fit(train_x, train_y, epochs=2, validation_split=0.2)
+    assert clf.predict(test_x).shape == (len(train_x), 1)
```
fchollet: Only checking the output shape in each case seems like it wouldn't be quite sufficient. We should check things like how features ended up being used, what's the accuracy, etc...

tests/test_task.py
```diff
@@ -0,0 +1,67 @@
...
+        id_to_word[i] for i in sentence), x_test))
+    x_train = np.array(x_train, dtype=np.str)
+    x_test = np.array(x_test, dtype=np.str)
+    return (x_train, y_train), (x_test, y_test)
+
+
+def test_text_classifier(tmp_dir):
+    (train_x, train_y), (test_x, test_y) = imdb_raw()
+    clf = ak.TextClassifier(directory=tmp_dir, max_trials=2)
+    clf.fit(train_x, train_y, epochs=2, validation_split=0.2)
```
fchollet: Would it be relevant to have an `evaluate` method as well? e.g. a standard practice I would expect is to have 2 validation sets, one used for search during training, and one used at the end as the actual validation set (to prevent validation overfitting).

tests/test_task.py
jhfjhfj1: An issue is created for this comment will be address in the future.

tests/test_task.py
jhfjhfj1: An issue is created for this comment will be address in the future.

PR #693, opened 2019-07-02
Update split validation data. Use tf dataset for preprocessing

### LINKED ISSUE #690: Should we handle validation data or leave it to Keras?

jhfjhfj1:
resolves #690


autokeras/auto/tuner.py
```diff
@@ -1,20 +1,35 @@
...
+
 
 class AutoTuner(kerastuner.Tuner):
     """Modified KerasTuner base class to include preprocessing layers."""
 
     def run_trial(self, trial, hp, fit_args, fit_kwargs):
         """Preprocess the x and y before calling the base run_trial."""
-        fit_kwargs.update(dict(zip(tf.keras.Model.fit.__code__.co_varnames,
+        new_fit_kwargs = copy.copy(fit_kwargs)
+        new_fit_kwargs.update(dict(zip(tf.keras.Model.fit.__code__.co_varnames,
```
fchollet: This does not seem like a reliable way to obtain the names of the positional arguments of `fit`, prefer using `getfullargspec`

autokeras/hypermodel/processor.py
```diff
@@ -80,23 +81,33 @@ def __init__(self, **kwargs):
...
-        self.std = np.std(data,
-                          axis=axis,
-                          keepdims=True).flatten()
+        shape = utils.dataset_shape(data)
+        axis = tuple(range(len(shape) - 1))
+
+        def sum_up(old_state, new_elem):
+            return old_state + new_elem
+
+        def sum_up_sqaure(old_state, new_elem):
```
fchollet: square

autokeras/utils.py
```diff
@@ -0,0 +1,100 @@
...
+def dataset_shape(dataset):
+    return tf.compat.v1.data.get_output_shapes(dataset)
+
+
+def input_list_to_dataset(x):
+    new_x = []
+    for temp_x in x:
+        if isinstance(temp_x, np.ndarray):
+            new_x.append(tf.data.Dataset.from_tensor_slices(temp_x))
+    return new_x
```
fchollet: This returns a list of datasets, but really it should return one dataset that yields a list, i.e. 

`tf.data.Dataset.from_tensor_slices(x)` where `x` is an iterable of numpy arrays.

autokeras/utils.py
```diff
@@ -0,0 +1,100 @@
...
+        y = format_inputs(y, 'train_y')
+        y = input_list_to_dataset(y)
+    if validation_data:
+        x_val, y_val = validation_data
+        x_val = format_inputs(x_val, 'x_val')
+        y_val = format_inputs(y_val, 'y_val')
+        x_val = input_list_to_dataset(x_val)
+        y_val = input_list_to_dataset(y_val)
+        validation_data = x_val, y_val
+    return x, y, validation_data
```
fchollet: This returns a tuple of lists of datasets, but really it should return 2 datasets, one for the training data (that yields x, y) and one for the validation data (that yields x_val, y_val).

autokeras/auto/auto_model.py
```diff
@@ -191,13 +192,30 @@ def fit(self,
...
+            validation_split: Float between 0 and 1.
+                Fraction of the training data to be used as validation data.
+                The model will set apart this fraction of the training data,
+                will not train on it, and will evaluate
+                the loss and any model metrics
+                on this data at the end of each epoch.
+                The validation data is selected from the last samples
+                in the `x` and `y` data provided, before shuffling. This argument is
+                not supported when `x` is a dataset, dataset iterator, generator or
+               `keras.utils.Sequence` instance.
```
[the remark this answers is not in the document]
jhfjhfj1: Currently, we only support numpy.ndarray and tf.data.Dataset. I will remove the generators and Sequence from the docstring for now and add them back later if we will support them.

autokeras/utils.py
jhfjhfj1: If the inputs are all patched into one Dataset, they cannot be passed to different preprocessing blocks separately. Each preprocessor would get all the inputs, they have to know which one to use.

autokeras/utils.py
jhfjhfj1: I think what we really need is a list of datasets instead of one dataset.
I renamed the function to `inputs_to_datasets`.

jhfjhfj1:
Since the only issue remains is whether the feeding data to preprocessors should be a single Dataset or multiple dataset, I will merge the pull request to not blocking other pull requests. Since it will be a big change, I will pull request separately if needed.

PR #709, opened 2019-07-16
class ImageAugment

Davidsirui:
resolves #000


autokeras/hypermodel/processor.py
```diff
@@ -255,3 +255,140 @@ def output_types(self):
...
+        self.saturation_range = saturation_range
+        self.contrast_range = contrast_range
+        self.horizontal_flip = horizontal_flip
+        self.vertical_flip = vertical_flip
+        self.whether_translation = whether_translation
+        self.gaussian_noise = gaussian_noise
+        self._shape = None
+
+    @staticmethod
+    def __get_min_and_max(value, name):
```
jhfjhfj1: use one "_" instead of two.

autokeras/hypermodel/processor.py
```diff
@@ -255,3 +255,140 @@ def output_types(self):
...
+        else:
+            min_value = 1. - value
+            max_value = 1. + value
+        return min_value, max_value
+
+    def transform(self, x):
+        self._shape = x.shape
+        rotation_range = self.rotation_range
+        if rotation_range is None:
+            rotation_range = self._hp.Choice('rotation_range')
```
jhfjhfj1: Please look into the implementation of hp here: https://github.com/keras-team/keras-tuner/blob/master/kerastuner/engine/hyperparameters.py
You may take a look at the block.py for the usage of the hp.
We need the name, values, defaults all specified when we call it.
Besides Choice, we also have Range, Linear and Fixed. Please select the best one for each hp.

tests/test_processor.py
```diff
@@ -61,3 +61,25 @@ def map_func(x):
...
+                                       whether_random_crop=True,
+                                       brightness_range=0.3,
+                                       saturation_range=0.4,
+                                       contrast_range=0.3,
+                                       horizontal_flip=True,
+                                       vertical_flip=True,
+                                       whether_translation=True,
+                                       gaussian_noise=True)
+    dataset = tf.data.Dataset.from_tensor_slices(raw_images)
+    #augmenter.set_hp(kerastuner.HyperParameter())
```
jhfjhfj1: After you address the change of hp usage. This line should be uncomment. And it should work.

autokeras/hypermodel/preprocessor.py
```diff
@@ -264,3 +265,176 @@ def set_state(self, state):
...
+        rotation_range: Int. The value can only be 0, 90, or 180.
+            Degree range for random rotations. Default to 180.
+        random_crop: Boolean. Whether crop the image randomly. Default to True.
+        brightness_range: Positive int.
+            Serve as 'max_delta' in tf.image.random_brightness. Default to 0.5.
+        saturation_range: Positive int or Tuple.
+            Serve as the bound of the tf.image.random_saturation. Default to 0.5.
+        contrast_range: Positive int or Tuple.
+            Serve as the bound of the tf.image.random_contrast. Default to 0.5.
+        translation: Boolean. Whether translate the image.
```
fchollet: 'Whether to" (below as well)

PR #733, opened 2019-08-12
MetaModel restricts ImageBlock search space based on dataset

jhfjhfj1:
resolves #000


autokeras/hypermodel/graph.py
```diff
@@ -388,3 +390,10 @@ def _get_block(self, name):
             if block.name == name:
                 return block
         return None
+
+    def set_hps(self, hps):
+        self._hps = hps
+
+    def _init_hps(self, hp):
```
fchollet: Why have this step at all? Why not pass hp instances to `set_hps`?

jhfjhfj1: The hp instance for build() function is totally created and controlled by Tuner class. We can not replace that hp with our hp instance. We can only change that hp.

This solution restricts the space of hp only by accessing the public functions, e.g. Choice, of HyperParameters class.  

On the other hand, an alternative would be we pass a hp instance and merge it into the hp provided by Tuner. However, it reaches into the inner representation of the space, which might not be as good as the first solution.

fchollet: I would actually recommend using the object representation (e.g. `hp_module.Int('int', 1, 10)`) and editing the target `HyperParameters` instance by adding entries in `hp.values` and `hp.space`. It should be simpler and cleaner, and achieve the same result.

PR #741, opened 2019-08-20
Enabled a new rule with flake8.

gabrieldemarmiesse:
Also enable flake8 test on the whole repo.

1) We enable flake8 on the whole repo, after all, this is a project in development, let's make sure that the examples are easy to read for people who want to understand the API.

2) We enable a new flake8 rule. Don't use undefined variables.

Note that since there is no numpy backend in this repo, we can remove the corresponding code in the docs generation file.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -23,8 +23,6 @@ ignore =
     N806
     # blank line at end of file
     W391
-    # undefined name
-    F821
 
 
 max-line-length = 85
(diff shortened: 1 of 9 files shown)


gabrieldemarmiesse:
Codacy needs a private token to work. For this reason, it's impossible to use it in forks. I've added a check to avoid using codacy if we're on a fork in travis CI.

.flake8
```diff
@@ -23,8 +23,6 @@ ignore =
     N806
     # blank line at end of file
```
jhfjhfj1: Maybe we can enable blank line check, too?

gabrieldemarmiesse: Sure, it can be done in a separate PR. The goal is to remove all exceptions as long as it doesn't conflict with a major design decision.

PR #744, opened 2019-08-21
[MRG] Add default early stop during the search

chengchengeasy:
### Pull Request Status 
Read to merged

Add default early stop during the search


autokeras/tuner.py
```diff
@@ -67,6 +71,27 @@ def load_trial(self, trial):
...
+            raise ValueError(
+                'All callbacks used during a search '
+                'should be deep-copyable (since they are '
+                'reused across executions). '
+                'It is not possible to do `copy.deepcopy(%s)`' %
+                (callbacks,))
+
+        if not [callback for callback in callbacks
+                if isinstance(callback, tf.keras.callbacks.EarlyStopping)]:
+            callbacks.append(tf.keras.callbacks.EarlyStopping(patience=30))
```
fchollet: 30 seems like a fairly arbitrary config. At least there should be a comment to explain this choice. Perhaps it should be made configurable by the user?

jhfjhfj1: I will add the comment. 30 is just based on my experience.

The user should pass their own earlystopping in the callbacks if they would like to configure it.

PR #746, opened 2019-08-22
Feature engineering

### LINKED ISSUE #681: Support 2D image inputs e.g. 28*28, reshape to 28*28*1

yufei-12:
1. Added FeatureEngineering preprocessor.
2. Added StructuredDataClassifier and Regressor.
3. Added StructuredDataBlock.
4. The users can specify column_names and column_types in StructuredDataInput(). The column_names can also be inferred from csv file or pd.DataFrame.
5. Add fit() and transform() function to Head and Input class. The subclasses override the functions to transform the data input by the user to tf.data.Dataset. It resolves #681.
6. The GraphHyperModel class is refactored into a base class and two subclasses. The two subclasses correspond to graph with HyperBlock (GraphHyperModel) and without HyperBlock (HyperBuiltGraphHyperModel).
7. Support compile() function of GraphHyperModel. It passes the information in the graph around, from block to block, e.g., passes the column_names from StructuredDataInput() to FeatureEngineering(), passes the head type to LightGBMBlock() to let it know it is a regression or classification.


autokeras/hypermodel/preprocessor.py
```diff
@@ -673,3 +676,233 @@ def set_config(self, config):
...
+
+class FeatureEngineering(Preprocessor):
+    """A preprocessor block does feature engineering for the data.
+
+    # Arguments
+        column_types: A list of strings. The length of the list should be the same
+            as the number of columns of the data. The strings in the list are
+            specifying the types of the columns. They should either be 'numerical'
+            or 'categorical'.
+        max_columns: Int. The maximum number of columns after feature engineering.
```
fchollet: The meaning of this argument isn't entirely clear. Is this about feature selection?

autokeras/hypermodel/preprocessor.py
```diff
@@ -673,3 +676,233 @@ def set_config(self, config):
...
+        # LabelEncoding
+        for col_index in self.categorical_col:
+            key = str(x[col_index])
+            try:
+                x[col_index] = self.label_encoders[col_index].transform(key)
+            except KeyError:
+                x[col_index] = -1
+        return np.hstack((x, np.array(new_values)))
+
+    def _impute(self, x):
```
fchollet: It would be more readable to call it e.g. `fill_missing` since `impute` looks somewhat like `input` and isn't a very common word

autokeras/hypermodel/preprocessor.py
fchollet: The meaning of this argument isn't entirely clear. Is this about feature selection? E.g. the max dimensionality of the output space?

autokeras/hypermodel/preprocessor.py
```diff
@@ -673,3 +676,233 @@ def set_config(self, config):
...
+
+    def output_types(self):
+        return (tf.float64,)
+
+    @property
+    def output_shape(self):
+        return self._shape
+
+    def get_weights(self):
+        return {'_shape': self._shape,
```
fchollet: Config dicts should be considered as part of the public API. As such they should not contain private symbol. Consider removing the underscore here.

autokeras/hypermodel/preprocessor.py
```diff
@@ -673,3 +676,233 @@ def set_config(self, config):
...
+        for col_index1, col_index2 in self.categorical_categorical.keys():
+            key = (str(x[col_index1]), str(x[col_index2]))
+            self.categorical_categorical[(col_index1, col_index2)][key] += 1
+
+        for num_col_index, cat_col_index in self.numerical_categorical.keys():
+            key = str(x[cat_col_index])
+            v = x[num_col_index]
+            self.numerical_categorical[(num_col_index, cat_col_index)][key] += v
+
+    def transform(self, x, fit=False):
```
fchollet: What is the type of `x`?

autokeras/hypermodel/preprocessor.py
```diff
@@ -673,3 +676,233 @@ def set_config(self, config):
...
+
+def return_zero():
+    return 0
+
+
+class FeatureEngineering(Preprocessor):
+    """A preprocessor block does feature engineering for the data.
+
+    # Arguments
+        column_types: A list of strings. The length of the list should be the same
```
fchollet: Passing a list that has the same length as the number of features in the data isn't a great API. It is very error prone.

Could it be feasible to pass a dict mapping column names (e.g. from a header) to their type? If a name is missing, its type could be automatically inferred from a sample data (which works in most cases): if all values in the sample data encode floats/ints, treat as numerical, else treat as categorical.

tests/hypermodel/test_preprocessor.py
```diff
@@ -124,6 +125,50 @@ def map_func(x):
     assert isinstance(new_dataset, tf.data.Dataset)
 
 
+def test_feature_engineering():
+    data = structured_data()
+    dataset = tf.data.Dataset.from_tensor_slices(data)
+    feature = preprocessor.FeatureEngineering([
+        'categorical', 'categorical', 'numerical', 'numerical', 'categorical',
+        'categorical', 'categorical', 'categorical'])
```
fchollet: This API is error prone and does not scale well to hundreds of features.

A better API: types are inferred most of the time (so the user doesn't specify anything). In cases where the user wants to treat e.g. an int column as categorical, the user passes `{column_name: 'categorical'}`

autokeras/hypermodel/preprocessor.py
jhfjhfj1: Yes, it is the maximum number of columns that the output can have. Do you think max_output_dim would be better?

autokeras/hypermodel/preprocessor.py
jhfjhfj1: eager tensor, I expect. This is in the map function of the dataset.

autokeras/hypermodel/preprocessor.py
jhfjhfj1: Does tf.data.Dataset comes with column names or header? If so we can easily change, otherwise it would be hard for the pandas dataframe to reach in from fit function all the way to this block.

tests/hypermodel/test_preprocessor.py
jhfjhfj1: Yes, we can refactor a little bit to infer the column types. however, does tf.data.Dataset has a header of column names?

fchollet: It depends on what you intend to be the input to the pipeline.

- A raw CSV file? In that case we could use the header.
- A tf.data Dataset created with `make_csv_dataset`? In that case it yields a dictionary and therefore contains column names. [Ref](https://www.tensorflow.org/beta/tutorials/load_data/csv)
- A Dataset or array structured as a list of values? In that case it won't contain column name information, and it would have to be passed via a different mechanism (e.g. a `header` argument in the constructor, which would be a list of strings)

I think it would be reasonable to require a dict input (dict of arrays or Dataset that yield dicts), in which case the column name information is part of the data. Would this choice create any issues?

autokeras/hypermodel/preprocessor.py
fchollet: There are 2 forms of feature selection one might be doing:

- 1) Removing entire columns that don't seem informative (either numerical or categorical columns)
- 2) Restricting the set of values taken into account for a categorical column

Both of these affect the output dimension. For instance if you have a categorical column with 100 occurrences of value "a", 150 of value "b" and 3 of value "c", you might drop "c" and encode the feature as a 2D binary vector (or even 1D).

"max_columns" makes it sound like you're only going to do 1). Is that the case?

jhfjhfj1: We are not removing columns but adding columns. 
So we are not doing 1) but something similar to 2) but not removing the original column but adding the new columns.
At the current stage, we expect max_columns should be larger than the total number of columns, so we would generate more features to reach that limit.

tests/hypermodel/test_preprocessor.py
jhfjhfj1: By dict input, do you mean each column name is a key, where the values are arrays (for the dict of array case) and the values are single values while iterating the Dataset?

I think this solution is good. It means we will support different types of input for different type of input nodes. I don't think we can support dict input for image input node.

I am intended to convert whatever input type from the user (csv, dataset, and etc) to a tf.Dataset with a header. Do you think it should be good?

fchollet: Yes, I think being able to accept the raw data from the user, in particular CSV files, is a good idea.

autokeras/hypermodel/preprocessor.py
```diff
@@ -673,3 +725,249 @@ def set_config(self, config):
...
+                    self.high_level_num_cat[pair][key] /= self.value_counters[
+                        cat_col_index1][key]
+
+        self.shape = (len(self.input_node.column_types)
+                      + len(self.high_level1_col)
+                      + len(self.high_level_cat_cat)
+                      + len(self.high_level_num_cat),)
+
+    def output_types(self):
+        return (tf.float64,)
```
jhfjhfj1: float32

autokeras/meta_model.py
```diff
@@ -156,7 +158,67 @@ def assemble(self, input_node):
 
 
 class StructuredDataAssembler(Assembler):
-    pass
+    """Assembler for structured data. which infers the column types for the data.
```
jhfjhfj1: update docstring

autokeras/utils.py
```diff
@@ -50,25 +50,18 @@ def validate_num_inputs(inputs, num):
...
-    y_val = []
-    for temp_x in x:
-        x_train.append(temp_x[:-validation_set_size])
-        x_val.append(temp_x[-validation_set_size:])
-    for temp_y in y:
-        y_train.append(temp_y[:-validation_set_size])
-        y_val.append(temp_y[-validation_set_size:])
-
-    return (x_train, y_train), (x_val, y_val)
+def split_dataset(dataset, validation_split):
```
jhfjhfj1: docstring

autokeras/hypermodel/head.py
```diff
@@ -142,8 +153,13 @@ def __init__(self,
                  output_dim=None,
                  loss=None,
                  metrics=None,
-                 dropout_rate=None):
-        super().__init__(loss=loss, metrics=metrics)
+                 dropout_rate=None,
+                 output_shape=None,
```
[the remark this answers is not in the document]
jhfjhfj1: removed output_shape from subclasses.

autokeras/hypermodel/head.py
```diff
@@ -78,13 +70,19 @@ def __init__(self,
                  multi_label=False,
                  loss=None,
                  metrics=None,
-                 dropout_rate=None):
-        super().__init__(loss=loss, metrics=metrics)
+                 dropout_rate=None,
+                 output_shape=None,
```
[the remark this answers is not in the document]
jhfjhfj1: removed output_shape from subclasses.

autokeras/__init__.py
```diff
@@ -12,11 +12,10 @@
 from autokeras.hypermodel.head import RegressionHead
```
jhfjhfj1: add lightgbm block

autokeras/hypermodel/graph.py
```diff
@@ -27,69 +27,31 @@ def __init__(self, inputs, outputs, name=None):
...
-        model = tf.keras.Model(
-            [real_nodes[self._node_to_id[input_node]] for input_node in
-             self._model_inputs],
-            [real_nodes[self._node_to_id[output_node]] for output_node in
-             self.outputs])
+    def set_hps(self, hps):
+        """Set the hyperparameters to constrain the search space.
 
-        return self._compiled(hp, model)
+        Args:
```
jhfjhfj1: Change the docstring format

autokeras/hypermodel/graph.py
```diff
@@ -217,23 +171,74 @@ def _add_node(self, input_node):
...
+
+class HyperBuiltGraphHyperModel(GraphHyperModelBase):
+    """A HyperModel based on connected Blocks.
+
+    It is used by GraphHyperModel. GraphHyperModel's hyper_build function produces
+    an instance of HyperBuiltGraphHyperModel, which can be directly built into Keras
+    Model.
+
+    # Arguments
+        inputs: A list of input node(s) for the GraphHyperModel.
```
jhfjhfj1: for HyperBuiltGraphHyperModel

autokeras/hypermodel/graph.py
```diff
@@ -11,8 +11,8 @@
 from autokeras.hypermodel import preprocessor
 
 
-class GraphHyperModel(kerastuner.HyperModel):
-    """A HyperModel based on connected HyperBlocks.
+class GraphHyperModelBase(kerastuner.HyperModel):
+    """A HyperModel based on connected Blocks or HyperBlocks.
 
     # Arguments
         inputs: A list of input node(s) for the GraphHyperModel.
```
jhfjhfj1: delete "for the GraphHyperModel"

autokeras/hypermodel/graph.py
```diff
@@ -407,12 +377,85 @@ def _get_block(self, name):
...
+    # these two functions in all blocks.
+    block = copy.copy(old_block)
+    block.clear_nodes()
+    if isinstance(block, preprocessor.Preprocessor):
+        block.clear_weights()
+    return block
+
+
+class GraphHyperModel(GraphHyperModelBase):
+
```
jhfjhfj1: Add docstring

autokeras/hypermodel/node.py
```diff
@@ -27,19 +33,86 @@ class TextNode(Node):
...
+        if isinstance(x, np.ndarray):
+            if x.dtype == np.float64:
+                x = x.astype(np.float32)
+            return tf.data.Dataset.from_tensor_slices(x)
+        raise TypeError('Unsupported type {type} for '
+                        '{name}.'.format(type=type(x),
+                                         name=self.__class__.__name__))
 
 
 class ImageInput(Input):
```
jhfjhfj1: docstring

autokeras/hypermodel/node.py
```diff
@@ -27,19 +33,86 @@ class TextNode(Node):
...
-    pass
+
+    def transform(self, x):
+        if isinstance(x, np.ndarray):
+            if len(x.shape) == 3:
+                x = x.reshape(-1, 1)
+        return super().transform(x)
 
 
 class TextInput(Input, TextNode):
```
jhfjhfj1: docstring

autokeras/hypermodel/node.py
```diff
@@ -27,19 +33,86 @@ class TextNode(Node):
...
     pass
 
 
-class StructuredInput(Input):
-    pass
+class StructuredDataInput(Input):
+    """Input node for structured data.
```
jhfjhfj1: docstring describing what type of data can be passed to this node.

autokeras/auto_model.py
```diff
@@ -169,34 +216,14 @@ def predict(self, x, batch_size=32, **kwargs):
...
         y = nest.flatten(y)
-        if not self._label_encoders:
-            return y
         new_y = []
-        for temp_y, label_encoder in zip(y, self._label_encoders):
-            if label_encoder:
-                new_y.append(label_encoder.decode(temp_y))
-            else:
-                new_y.append(temp_y)
+        for temp_y, head_block in zip(y, self.heads):
```
jhfjhfj1: Add inverse_transform() function to Head.

autokeras/task.py
```diff
@@ -167,3 +167,176 @@ def __init__(self,
...
+            validation_split=0,
+            validation_data=None,
+            **kwargs):
+        """Search for the best model and hyperparameters for the task.
+
+        # Arguments
+            x: String, numpy.ndarray, pandas.DataFrame or tensorflow.Dataset.
+                Training data x. If the data is from a csv file, it should be a
+                string specifying the path of the csv file of the training data.
+            y: String, numpy.ndarray, or tensorflow.Dataset. Training data y.
```
fchollet: - How is it possible for `y` to be a dataset?
- Wouldn't "string" also be valid if `x` is a dataframe? (y = name of target column)

It seems to me the possibilities are:

- x = dataset, y = None
- x = path to csv file, y = column name
- x = dataframe, y = column name
- x = np, y = np

Right?

autokeras/task.py
```diff
@@ -167,3 +167,176 @@ def __init__(self,
...
+            file or the pandas.DataFrame.
+        column_types: Dict. The keys are the column names. The values should either
+            be 'numerical' or 'categorical', indicating the type of that column.
+            Defaults to None. If not None, the column_names need to be specified.
+            If None, it will be inferred from the data.
+        num_classes: Int. Defaults to None. If None, it will infer from the data.
+        multi_label: Boolean. Defaults to False.
+        loss: A Keras loss function. Defaults to None. If None, the loss will be
+            inferred from the AutoModel.
+        metrics: A list of Keras metrics. Defaults to None. If None, the metrics will
```
fchollet: What does it imply for the metrics to be inferred? Is the list of default metrics known in advance? (e.g. always Accuracy?)

autokeras/task.py
```diff
@@ -167,3 +167,176 @@ def __init__(self,
...
+            length of the list should be equal to the number of columns of the data.
+            Defaults to None. If None, it will obtained from the header of the csv
+            file or the pandas.DataFrame.
+        column_types: Dict. The keys are the column names. The values should either
+            be 'numerical' or 'categorical', indicating the type of that column.
+            Defaults to None. If not None, the column_names need to be specified.
+            If None, it will be inferred from the data.
+        num_classes: Int. Defaults to None. If None, it will infer from the data.
+        multi_label: Boolean. Defaults to False.
+        loss: A Keras loss function. Defaults to None. If None, the loss will be
```
fchollet: What does it imply for the loss to be inferred? Is the default loss known known in advance?

autokeras/task.py
jhfjhfj1: 1. If x and y are both datasets, we would zip them together.
2. If x is a dataframe, we are not supporting the column name string of y.
3. Those are the typical ways of passing x and y. The user can also do something like: x=dataframe, y = np.

autokeras/task.py
jhfjhfj1: It would always be Accuracy for classification and always MSE for regression.
Yeah, the docstrings are confusing. I will change it.

autokeras/task.py
jhfjhfj1: It would always be (categorical or binary) crossentropy for classification and MSE for regression

jhfjhfj1: It is always (categorical or binary) crossentropy for classification and MSE for regression.

PR #750, opened 2019-08-24
Lgbm model

Davidsirui:
resolves #000

### DIFF
--- a/autokeras/hypermodel/preprocessor.py
+++ b/autokeras/hypermodel/preprocessor.py
@@ -218,7 +218,7 @@ class TextToNgramVector(Preprocessor):
         self._texts = []
         self._shape = None
 
-    def update(self, x):
+    def update(self, x, y=None):
         # TODO: Implement a sequential version fit for both
         #  TfidfVectorizer and SelectKBest
         self._texts.append(nest.flatten(x)[0].numpy().decode('utf-8'))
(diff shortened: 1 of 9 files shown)


autokeras/hypermodel/preprocessor.py
```diff
@@ -328,6 +330,96 @@ def clear_weights(self):
         self._shape = None
 
 
+class LgbmClassifier(Preprocessor):
```
fchollet: LightGBMClassifier would be a more explicit name

autokeras/hypermodel/preprocessor.py
```diff
@@ -328,6 +330,96 @@ def clear_weights(self):
...
+    """Collect data, train and test the LightGBM for classification task.
+
+    Input data are np.array etc. np.random.rand(example_number, feature_number).
+    Input labels are encoded labels in np.array form
+    etc. np.array([[1, 0, 0, 0, 0, 0, 0, 0, 0, 0],
+    [1, 0, 0, 0, 0, 0, 0, 0, 0, 0]]).
+    Outputs are predicted encoded labels in np.array form.
+
+    The instance of this LgbmClassifier class must be followed by
+    an IdentityBlock and an EmptyHead.
```
fchollet: Why? What is the purpose of this setup?

tests/hypermodel/test_preprocessor.py
```diff
@@ -113,3 +120,36 @@ def map_func(x):
...
+                        [0, 0, 0, 0, 0, 0, 0, 0, 0, 1],
+                        [0, 0, 0, 0, 0, 0, 0, 1, 0, 0],
+                        [1, 0, 0, 0, 0, 0, 0, 0, 0, 0],
+                        [0, 0, 0, 1, 0, 0, 0, 0, 0, 0],
+                        [0, 0, 0, 1, 0, 0, 0, 0, 0, 0]])
+
+    input_node = ak.Input()
+    output_node = input_node
+    output_node = ak.hypermodel.preprocessor.LgbmClassifier()(output_node)
+    output_node = ak.IdentityBlock()(output_node)
```
fchollet: This is a strange setup. We should not require users to use objects that don't have a clear purpose like IdentityBlock or EmptyHead.

PR #751, opened 2019-08-24
Weights

jhfjhfj1:
resolves #000

### DIFF
--- a/autokeras/hypermodel/preprocessor.py
+++ b/autokeras/hypermodel/preprocessor.py
@@ -150,7 +150,7 @@ class Normalization(Preprocessor):
     def output_shape(self):
         return self.inputs[0].shape
 
-    def get_config(self):
+    def get_weights(self):
         return {'sum': self.sum,
                 'square_sum': self.square_sum,
                 'count': self.count,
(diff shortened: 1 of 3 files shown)


autokeras/hypermodel/preprocessor.py
```diff
@@ -141,21 +171,29 @@ def output_types(self):
...
-        self.mean = state['mean']
-        self.std = state['std']
-        self._shape = state['_shape']
+    def set_weights(self, weights):
+        self.sum = weights['sum']
+        self.square_sum = weights['square_sum']
+        self.count = weights['count']
+        self.mean = weights['mean']
+        self.std = weights['std']
+        self._shape = weights['_shape']
```
fchollet: What is `_shape`? Note that state elements that end up in saved files should generally not be underscore-private.

autokeras/hypermodel/preprocessor.py
```diff
@@ -190,20 +228,28 @@ def output_types(self):
...
-                '_max_len': self._max_len,
+    def get_config(self):
+        return {'max_len': self.max_len}
+
+    def set_config(self, config):
+        self.max_len = config['max_len']
+
+    def get_weights(self):
+        return {'_max_len': self._max_len,
                 '_tokenizer': self._tokenizer}
```
fchollet: Same comment about tokenizer: should it be public?

PR #776, opened 2019-09-19
Partial column types

yufei-12:
This pull request adds the function of supporting partial column_types provided by user.


autokeras/hypermodel/preprocessor.py
```diff
@@ -8,6 +8,7 @@
 from sklearn.feature_extraction import text
 from tensorflow.python.util import nest
 
+import autokeras as ak
```
jhfjhfj1: remove this line

tests/hypermodel/test_preprocessor.py
```diff
@@ -129,6 +129,26 @@ def map_func(x):
     assert isinstance(new_dataset, tf.data.Dataset)
 
 
+column_names_for_tests = [
```
jhfjhfj1: avoid the duplication

autokeras/meta_model.py
```diff
@@ -218,6 +218,11 @@ def assemble(self, input_node):
         self.infer_column_types()
         if input_node.column_types is None:
             input_node.column_types = self.column_types
+        # partial column_types is provided
+        elif len(input_node.column_types) < len(input_node.column_names):
```
jhfjhfj1: I don't think we need this if. Just directly do the for loop.

PR #790, opened 2019-09-25
Final fit after early stopping

### LINKED ISSUE #779: Final fit after early stop

jhfjhfj1:
resolves #779 

### Pull Request Status 
In progress.

If epoch is not specified, use earlystopping to do it adaptively.
If epoch is specified, but earlystopping is not in the callbacks, we use earlystopping during the search to accelerate, then we do a final fit after we found the best model to finish the rest of the epochs.
Improved some tests. Use fixed seed during unit tests.

### DIFF
--- a/autokeras/hypermodel/preprocessor.py
+++ b/autokeras/hypermodel/preprocessor.py
@@ -337,12 +337,13 @@ class TextToNgramVector(Preprocessor):
 
 class LightGBMModel(Preprocessor):
     """The base class for LightGBMClassifier and LightGBMRegressor."""
-    def __init__(self, **kwargs):
+    def __init__(self, seed=None, **kwargs):
         super().__init__(**kwargs)
         self.data = []
         self.targets = []
         self._output_shape = None
         self.lgbm = None
+        self.seed = seed
 
     def transform(self, x, fit=False):
         """ Transform the data using well-trained LightGBM regressor.
@@ -406,7 +407,7 @@ class LightGBMClassifier(LightGBMModel):
 
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
-        self.lgbm = lgb.LGBMClassifier()
+        self.lgbm = lgb.LGBMClassifier(random_state=self.seed)
         self._one_hot_encoder = utils.OneHotEncoder()
 
     def update(self, x, y=None):
@@ -465,7 +466,7 @@ class LightGBMRegressor(LightGBMModel):
 
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
-        self.lgbm = lgb.LGBMRegressor()
+        self.lgbm = lgb.LGBMRegressor(random_state=self.seed)
 
     def update(self, x, y=None):
         """ Store the train data.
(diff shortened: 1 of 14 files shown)


autokeras/hypermodel/preprocessor.py
```diff
@@ -406,7 +407,7 @@ class LightGBMClassifier(LightGBMModel):
 
     def __init__(self, **kwargs):
```
fchollet: You may want to list the `seed` argument in this docstring above.

autokeras/hypermodel/preprocessor.py
```diff
@@ -465,7 +466,7 @@ class LightGBMRegressor(LightGBMModel):
...
```
fchollet: You may want to list the `seed` argument in this docstring above.

autokeras/tuner.py
```diff
@@ -72,27 +68,71 @@ def load_trial(self, trial):
...
+        fit_kwargs.setdefault('epochs', None)
+
+        # Insert early-stopping permanently when epochs is None.
+        callbacks = fit_kwargs['callbacks']
+        if fit_kwargs['epochs'] is None:
+            fit_kwargs['epochs'] = 1000
+            if not any([isinstance(callback, tf.keras.callbacks.EarlyStopping)
+                        for callback in callbacks]):
+                callbacks = callbacks + [
+                    tf.keras.callbacks.EarlyStopping(patience=30)]
```
fchollet: 30 seems somewhat excessive, it guarantees overfitting. ~10 seems more reasonable, no?

autokeras/task.py
```diff
@@ -203,6 +205,11 @@ def fit(self,
             y: String, numpy.ndarray, or tensorflow.Dataset. Training data y.
                 If the data is from a csv file, it should be a string corresponding
                 to the label column.
+            epochs: Int. The number of epochs to train each model during the search.
+                If unspecified, we would use epochs equal to 1000 and early stopping
```
fchollet: "If unspecified, by default we train for a maximum of 1000 epochs but we stop training if the validation loss stops improving for 30 epochs (unless you specified an `EarlyStopping` callback as part of the `callbacks` argument, in which case this callback will determine early stopping)."

autokeras/tuner.py
```diff
@@ -72,27 +68,71 @@ def load_trial(self, trial):
...
+        fit_kwargs['callbacks'] = callbacks
+
+        # Insert early-stopping temporarily for the search.
```
fchollet: I don't understand the logic here. `final_fit` is going to be True IFF there is no EarlyStopping callback. But the only case where there is no EarlyStopping callback is if `fit_kwargs['epochs']` is set by the user *and* the user did not pass an EarlyStopping callback.

Please clarify what the intent is, and why.

jhfjhfj1: Yes, this is what we want. If the user specified the epochs and did not put EarlyStopping in the callbacks, it means the user wants the model to be trained to the specific epochs. In this case, we need the final fit to fully train the model, and will not use early stopping in the final fit.

This temporary early stopping is just for accelerating the search. It is not related to the case which user did not specify the epochs.

For the case of user not specifying epochs or user specified epochs and put earlystopping in  the callbacks, each trial would be run exactly according to the user specifications. So we don't need the final fit.

fchollet: I would suggest refactoring this part of the code to make the logic easier to follow (e.g. using an enumeration of cases), and adding comments.

PR #792, opened 2019-09-25
Docstring of time series forecaster

yufei-12:
This pull request adds the docstring of time-series data forecaster.

### DIFF
--- a/autokeras/task.py
+++ b/autokeras/task.py
@@ -481,13 +481,14 @@ class TimeSeriesForecaster(auto_model.AutoModel):
             be 'numerical' or 'categorical', indicating the type of that column.
             Defaults to None. If not None, the column_names need to be specified.
             If None, it will be inferred from the data.
-        lookback: Int. The number of time steps to used before forecasting time step.
-            If unspecified, it will be tuned automatically.
-        predict_from: Int. The starting point of the forecast. The number of time
-            steps after the training data. If unspecified, it will be tuned
+        lookback: Int. The range of history steps to consider for each prediction.
+            For example, if lookback=n, the data from step (i - n) to (i - 1) is used
+            to predict the value of step i. If unspecified, it will be tuned
             automatically.
-        predict_until: Int. The end point of the forecast. The number of time steps
-            after the training data. If unspecified, it will be tuned automatically.
+        predict_from: Int. The starting point of the forecast. The number of steps
+            after the last step in training data. Defaults to 1.
+        predict_until: Int. The end point of the forecast. The number of steps
+            after the last step in training data. Defaults to 10.
         loss: A Keras loss function. Defaults to use 'mean_squared_error'.
         metrics: A list of Keras metrics. Defaults to use 'mean_squared_error'.
         name: String. The name of the AutoModel. Defaults to
--- a/autokeras/task.py
+++ b/autokeras/task.py
@@ -488,6 +488,7 @@ class TimeSeriesForecaster(auto_model.AutoModel):
             automatically.
         predict_until: Int. The end point of the forecast. The number of time steps
             after the training data. If unspecified, it will be tuned automatically.
+        loss: A Keras loss function. Defaults to use 'mean_squared_error'.
         metrics: A list of Keras metrics. Defaults to use 'mean_squared_error'.
         name: String. The name of the AutoModel. Defaults to
             'time_series_forecaster'.
@@ -505,6 +506,7 @@ class TimeSeriesForecaster(auto_model.AutoModel):
                  lookback=None,
                  predict_from=1,
                  predict_until=10,
+                 loss='mean_squared_error',
                  metrics=None,
                  name='time_series_forecaster',
                  max_trials=100,
--- a/autokeras/task.py
+++ b/autokeras/task.py
@@ -481,11 +481,13 @@ class TimeSeriesForecaster(auto_model.AutoModel):
             be 'numerical' or 'categorical', indicating the type of that column.
             Defaults to None. If not None, the column_names need to be specified.
             If None, it will be inferred from the data.
-        in_steps: Int. Number of lag observations as input (x). Values may be between
-            [1..len(data)]. Defaults to 1. If unspecified, it will be tuned
+        lookback: Int. The number of time steps to used before forecasting time step.
+            If unspecified, it will be tuned automatically.
+        predict_from: Int. The starting point of the forecast. The number of time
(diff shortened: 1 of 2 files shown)


autokeras/task.py
```diff
@@ -333,3 +333,95 @@ def __init__(self,
...
+        # Arguments
+            x: String, numpy.ndarray, pandas.DataFrame or tensorflow.Dataset.
+                Training data x. If the data is from a csv file, it should be a
+                string specifying the path of the csv file of the training data.
+            y: A list of strings, numpy.ndarray, pandas.DataFrame or
+                tensorflow.Dataset. Training data y.
+                If the data is from a csv file, it should be a list of string(s)
+                specifying the name(s) of the column(s) need to be forecasted.
+                If it is multivariate forecasting, y should contain more than one
+                column of data.
```
jhfjhfj1: y should be a list of more than one column names.

autokeras/task.py
```diff
@@ -333,3 +333,95 @@ def __init__(self,
...
+            validation_split=0,
+            validation_data=None,
+            **kwargs):
+        """Search for the best model and hyperparameters for the task.
+        
+        # Arguments
+            x: String, numpy.ndarray, pandas.DataFrame or tensorflow.Dataset.
+                Training data x. If the data is from a csv file, it should be a
+                string specifying the path of the csv file of the training data.
+            y: A list of strings, numpy.ndarray, pandas.DataFrame or
```
jhfjhfj1: we also support single string, if it is univariate forecasting.

autokeras/task.py
```diff
@@ -333,3 +333,92 @@ def __init__(self,
...
+    # Arguments
+        column_names: A list of strings specifying the names of the columns. The
+            length of the list should be equal to the number of columns of the data.
+            Defaults to None. If None, it will obtained from the header of the csv
+            file or the pandas.DataFrame.
+        column_types: Dict. The keys are the column names. The values should either
+            be 'numerical' or 'categorical', indicating the type of that column.
+            Defaults to None. If not None, the column_names need to be specified.
+            If None, it will be inferred from the data.
+        in_steps: Int. Number of lag observations as input (x). Values may be between
```
fchollet: There are many ways to convert a time series into a fixed number of steps: in particular using strides and using exponential moving averages. Same for the output. This type of decision should be left to the AutoML algorithm.

The question that is relevant to the user is "how far in the past should we look" and "how far in the future should we predict". But just looking "up to 100 timesteps in the past" doesn't necessarily mean that the model will have an input length of 100 (e.g. you could be using strides, etc).

So I'd suggest a `lookback` argument (integer, maximum age of input data in timesteps) and a `predict_until` argument. We could also offer more flexibility and have e.g. `predict_from=1` and `predict_until=10`.

autokeras/task.py
```diff
@@ -333,3 +333,92 @@ def __init__(self,
...
+
+    def __init__(self,
+                 column_names=None,
+                 column_types=None,
+                 in_steps=1,
+                 out_steps=1,
+                 output_dim=None,
+                 loss=None,
+                 metrics=None,
+                 name='time_series_data_forecaster',
```
fchollet: The class name doesn't have `data` in it, just `time_series_forecaster`

autokeras/task.py
```diff
@@ -333,3 +333,92 @@ def __init__(self,
             max_trials=max_trials,
             directory=directory,
             seed=seed)
+
+
+class TimeSeriesForecaster(auto_model.AutoModel):
+    """AutoKeras time series data forecast class.
+
+    # Arguments
+        column_names: A list of strings specifying the names of the columns. The
```
fchollet: Remind me (this is the same API as the structure data classes): what happens if the data is already vectorized (e.g. Numpy) and we don't specify column names or column types? What if you just have raw data without semantic annotations?

autokeras/task.py
jhfjhfj1: 1. The user do not need to specify the column_names and column_types if the data are already vectorized. Actually, we only need the column_names if (1) the data are raw data and (2) column_names is not in the data and (3) the user is using the column_names to specify the types of the data in the column_types.

2. The meta_model would infer the column semantic types by looking into the values of the data.

autokeras/task.py
jhfjhfj1: 1. The user does not need to specify the column_names and column_types if the data are already vectorized. Actually, we only need the column_names if (1) the data are raw data and (2) column_names is not in the data and (3) the user is using the column_names to specify the types of the data in the column_types.

2. The meta_model would infer the column semantic types by looking into the values of the data.

autokeras/task.py
jhfjhfj1: Agree. So with `predict_from` the user can have a time gap between the past and the predicted? If it doesn't affect the AutoML choices, i.e. we would need to predict all the values in the time gap before the specified time range then cut out the gap part before output, I think we should leave it to the user to do the cut.

autokeras/task.py
```diff
@@ -467,3 +467,95 @@ def __init__(self,
...
+            file or the pandas.DataFrame.
+        column_types: Dict. The keys are the column names. The values should either
+            be 'numerical' or 'categorical', indicating the type of that column.
+            Defaults to None. If not None, the column_names need to be specified.
+            If None, it will be inferred from the data.
+        lookback: Int. The range of history steps to consider for each prediction.
+            For example, if lookback=n, the data in the range of [i - n, i - 1]
+            is used to predict the value of step i. If unspecified, it will be tuned
+            automatically.
+        predict_from: Int. The starting point of the forecast. The number of steps
```
fchollet: The descriptions for these 2 arguments could be improved a bit. In particular, "training data" is confusing as it normally refers to the training dataset (all training samples), but here it refers to the input steps in one sample (that isn't necessarily part of the training data).

E.g.

"Int. The starting point of the forecast for each sample (in number of steps) after the last time step in the input. If N is the last step in the input, then the first step of the predicted output will be `N + predict_from`. Defaults to 1 (which corresponds to starting the forecast immediately after the last step in the input)."

autokeras/task.py
```diff
@@ -467,3 +467,95 @@ def __init__(self,
...
+            after the last step in training data. Defaults to 1.
+        predict_until: Int. The end point of the forecast. The number of steps
+            after the last step in training data. Defaults to 10.
```
fchollet: "after the last step in the input."

PR #816, opened 2019-10-22
Concat validation set and training set for final_fit if using validation_split

jhfjhfj1:

resolves #000

### DIFF
--- a/autokeras/auto_model.py
+++ b/autokeras/auto_model.py
@@ -48,6 +48,7 @@ class AutoModel(object):
         self.directory = directory
         self.seed = seed
         self.hyper_graph = None
+        self._split_dataset = False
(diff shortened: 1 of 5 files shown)


autokeras/auto_model.py
```diff
@@ -139,6 +144,7 @@ def fit(self,
                     tf.keras.callbacks.EarlyStopping(patience=10)]
 
         self.tuner.search(x=dataset,
+                          concat=self._split_dataset,
```
fchollet: `search` is only called once per tuner instance, since the tuner is instantiated right above this line (L127). Would it make sense to have the argument be a Tuner constructor argument instead of being part of the `search` signature?

Also, `concat` as a name is not representative of what the argument means. Better to name it e.g. `fit_on_val_data`.

PR #853, opened 2019-12-13
Makefile for docker images. Docker with cuda and cudnn

Sette:

resolves #000

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -1,0 +1,58 @@
+# Using Autokeras via Docker
+
+This directory contains `Dockerfile` to make it easy to get up and running with
+Autokeras via [Docker](http://www.docker.com/).
+
+## Installing Docker
+
+General installation instructions are
+[on the Docker site](https://docs.docker.com/installation/), but we give some
+quick links here:
+
+* [OSX](https://docs.docker.com/installation/mac/): [docker toolbox](https://www.docker.com/toolbox)
+* [ubuntu](https://docs.docker.com/installation/ubuntulinux/)
+
+## Running the container
+
+We are using `Makefile` to simplify docker commands within make commands.
+
+Build the container and start a Jupyter Notebook
+
+    $ make notebook
+
+Build the container and start an iPython shell
+
+    $ make ipython
+
+Build the container and start a bash
+
+    $ make bash
+
+For GPU support install NVIDIA drivers (ideally latest) and
+[nvidia-docker](https://github.com/NVIDIA/nvidia-docker). Run using
+
+    $ make notebook GPU=0 # or [ipython, bash]
+
+Switch between Theano and TensorFlow
+
+    $ make notebook BACKEND=theano
+    $ make notebook BACKEND=tensorflow
+
+Mount a volume for external data sets
+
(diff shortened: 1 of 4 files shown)


gabrieldemarmiesse:
Could we just target tf.keras? It would make things much simpler. The multi-backend keras is deprecated.

Sette:
You're right. I will improve the dockerfile and remove the keras multi-beckend.

docker/README.md
```diff
@@ -0,0 +1,58 @@
...
+Build the container and start a bash
+
+    $ make bash
+
+For GPU support install NVIDIA drivers (ideally latest) and
+[nvidia-docker](https://github.com/NVIDIA/nvidia-docker). Run using
+
+    $ make notebook GPU=0 # or [ipython, bash]
+
+Switch between Theano and TensorFlow
```
jhfjhfj1: We are not supporting Theano backend for AutoKeras.

docker/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+    $ make notebook BACKEND=theano
```
jhfjhfj1: Same here.

docker/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+Mount a volume for external data sets
+
+    $ make DATA=~/mydata
+
+Prints all make tasks
+
+    $ make help
+
+You can change Theano parameters by editing `/docker/theanorc`.
```
jhfjhfj1: Same here.

haifeng-jin:
I am merging it.
I haven't test it yet, but it definitely more complete than the previous file.
We can fix it later if anyone finds any bug.

gabrieldemarmiesse:
@Sette tensorflow has ready to use docker images, with gpu enable and jupyter included. Do you think we could use that to reduce the size of the Dockerfile? 

https://hub.docker.com/r/tensorflow/tensorflow

haifeng-jin:
@gabrieldemarmiesse When we make sure the docker file is good, we may try the GitHub Package feature.
https://github.com/keras-team/autokeras/packages?package_type=Docker
What do you think?

gabrieldemarmiesse:
@haifeng-jin , you mean to host the produced docker image? We have the choice between GitHub package and DockerHub on this.

From [this thread](https://github.community/t5/GitHub-Actions/Use-docker-images-from-GitHub-Package-Registry/td-p/30407) it seems that dockerhub is more integrated into the docker cli than GitHub packages. (not a big surprise since it's the same company making both).

It seems that unless we have a strong reason not to, it would be preferable to push the images to DockerHub.

I hope I didn't misinterpret your question.

haifeng-jin:
@gabrieldemarmiesse Great! Then let's just use DockerHub. We may automate the push process with GitHub Actions.

gabrieldemarmiesse:
Ok sure thing. Beforehand, we should rewrite the Dockerfile to use the official Tensorflow image rather than the nvidia/cuda one. It's going to be less maintenance in the long run. I can work on that when I have the time.

haifeng-jin:
@gabrieldemarmiesse Awesome! Thank you

PR #875, opened 2019-12-28
Add the export model API

haifeng-jin:

resolves #000

### DIFF
--- a/autokeras/auto_model.py
+++ b/autokeras/auto_model.py
@@ -312,12 +312,12 @@ class AutoModel(object):
             self._process_xy(x, y))[0].batch(batch_size)
         return model.evaluate(data, **kwargs)
 
-    def export_keras_model(self):
+    def export_model(self):
         """Export the best Keras Model.
 
         # Returns
-            tf.keras.Model. The best model found during the search with trained
-            weights.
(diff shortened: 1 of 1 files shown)


autokeras/auto_model.py
```diff
@@ -311,3 +311,13 @@ def evaluate(self, x, y=None, batch_size=32, **kwargs):
         data = preprocess_graph.preprocess(
             self._process_xy(x, y))[0].batch(batch_size)
         return model.evaluate(data, **kwargs)
+
+    def export_keras_model(self):
+        """Export the best Keras Model.
+
+        # Returns
+            tf.keras.Model. The best model found during the search with trained
```
fchollet: "`tf.keras.Model` instance. The best model found during the search, loaded with trained weights."

autokeras/auto_model.py
```diff
@@ -311,3 +311,13 @@ def evaluate(self, x, y=None, batch_size=32, **kwargs):
         data = preprocess_graph.preprocess(
             self._process_xy(x, y))[0].batch(batch_size)
         return model.evaluate(data, **kwargs)
+
+    def export_keras_model(self):
```
fchollet: Would `export_model` suffice here?

PR #919, opened 2020-01-20
Simplified the dockerfile.

### LINKED ISSUE #913: Make the Dockerfile use the official TF image instead of nvidia/cuda
cf #853

gabrieldemarmiesse:
Fix #913

This is a good starting point. I'll see later on how we can build it and push it to Dockerhub.


gabrieldemarmiesse:
Yes indeed.

PR #922, opened 2020-01-21
Adopting the tf.keras preprocessing layers

### LINKED ISSUE #906: Adopt the text to int sequence or vector embeddings from preprocessing layers

haifeng-jin:

resolves #906 #907 #908 #909 #910


autokeras/adapters/input_adapter.py
```diff
@@ -99,30 +57,18 @@ def _check(self, x):
...
-            file or the pandas.DataFrame.
-        column_types: Dict. The keys are the column names. The values should either
-            be 'numerical' or 'categorical', indicating the type of that column.
-            Defaults to None. If not None, the column_names need to be specified.
-            If None, it will be inferred from the data. A column will be judged as
-            categorical if the number of different values is less than 5% of the
-            number of instances.
-    """
+    CATEGORICAL = 'categorical'
+    NUMERICAL = 'numerical'
```
fchollet: I would recommend not having global keys like this be class attributes. Instead, make it external constants.

autokeras/engine/block.py
```diff
@@ -0,0 +1,83 @@
...
+class Block(kerastuner.HyperModel, serializable.Serializable):
+    """The base class for different Block.
+
+    The Block can be connected together to build the search space
+    for an AutoModel. Notably, many args in the __init__ function are defaults to
+    be a tunable variable when not specified by the user.
+
+    # Arguments
+        name: String. The name of the block. If unspecified, it will be set
+        automatically with the class name.
```
fchollet: Need indent

autokeras/hypermodels/preprocessing.py
```diff
@@ -0,0 +1,202 @@
+import numpy as np
+from tensorflow.keras.layers.experimental import preprocessing
+from tensorflow.python.util import nest
+
+from autokeras import keras_layers
+from autokeras.adapters import input_adapter
+from autokeras.engine import block as block_module
+
+
+class Normalization(block_module.Block):
```
fchollet: Note that there is a Normalization layer in Keras.

autokeras/hypermodels/preprocessing.py
```diff
@@ -0,0 +1,202 @@
...
+        else:
+            raise ValueError('Expected {name} to be either a float between 0 and 1, '
+                             'or a tuple of 2 floats between 0 and 1, '
+                             'but got {value}'.format(name=name, value=value))
+
+    def build(self, hp, inputs=None):
+        return inputs
+
+
+class FeatureEncoding(block_module.Block):
```
fchollet: "FeatureEncoding" is very generic (everything in ML is a feature and needs encoding, since it usually doesn't start out as a numerical array). Perhaps CategoricalEncoding or CategoricalFeatureEncoding would be more specific? Or CategoricalToNumerical?

haifeng-jin: Right, `CategoricalToNumerical` would be more accurate and more consistent with other preprocessing blocks like `TextToNgramVector`.

PR #941, opened 2020-02-03
added structured data regression example

nocluebutalotofit:
Added structured data regression example with a quick and dirty performance comparison to lightGBM.


haifeng-jin:
@nocluebutalotofit Thanks for the pull request. We are currently focusing on Keras models (neural networks). When comparing with tree-based models, it may have its limitations.

The dataset is good. We will include this tutorial later.

Thank you!

Issue #918, opened 2020-01-20
Adding types information on the API surface.


gabrieldemarmiesse:
Adding types on the public API surface would allow us to do some runtime type checking later on and would allow user's IDE to have more info for static analysis. 

The functions/signatures to type are the ones listed here https://github.com/keras-team/autokeras/blob/master/autokeras/__init__.py

For the context, see #856 where I add some type information on a ImageClassifier.

This issue can be considered easy to solve (good first issue), it's just long to do because the public API surface is big.

I'll do it but some help is welcome :) make sure to make small pull requests.

gabrieldemarmiesse:
We're going to have a nice TODO list with #1051 . Then we'll be able to track easily what's left and we'll avoid regressions with it.

keremgocen:
how's this going? any more help needed?

shamiquekhan:
how's this going? any more help needed? . I can contibute .

PR #947, opened 2020-02-09
Added typing check.

gabrieldemarmiesse:
This is linked to #918 . We've had quite some success doing that with tensorflow addons to enforce a strict progression and ensure the TODO list is always up to date.

I assumed that all the functions/classes in the autokeras public API were in `autokeras/__init__.py` . Is that true or are them some of them somewhere else?

I didn't upload typedapi to pypi yet (not sure I'll ever do it since it's a single file) and adding a github link in the setup.py is very hard. This is why I added it in the CI directly.

`shell/lint.sh` and `shell/format.sh` are both telling me that everything is fine locally. This is very strange. It seems that the isort used locally gives me different results than the in CI. I used isort==4.3.21 . For the commit to pass the CI I had to manually go against what isort had done and change the imports myself.


haifeng-jin:
@gabrieldemarmiesse To use isort locally you need to pip uninstall autokeras and add the repo path to the python path so that isort won't treat autokeras as a third party lib installed from pypi, but the local import.

gabrieldemarmiesse:
@haifeng-jin thanks for the answer. In tensorflow/addons, we provide a pre-commit running with docker to ensure that it will always work for our users and stay up to date. Would that be something you'd like me to work on in autokeras? See https://github.com/tensorflow/addons/blob/master/tools/pre-commit.sh

haifeng-jin:
@gabrieldemarmiesse So this pre-commit.sh will be executed before each commit, right? If it is done with docker will it be very slow to make the commits?

gabrieldemarmiesse:
Not really. You can try it out by cloning tensorflow/addons if you want. There are two steps, first it builds the image with the tools in it, then runs the container to do all the formatting.

The first step takes ~3m in my computer the first time, then 0.2 seconds the second time (because of the docker cache)
The second step takes as long as using normal tools without docker. In our case 2-3 seconds since we have bazel format, clang-format, black and flake8 (soon isort too).

gabrieldemarmiesse:
> I assumed that all the functions/classes in the autokeras public API were in autokeras/__init__.py . Is that true or are them some of them somewhere else?

Ping about this question :)

haifeng-jin:
@gabrieldemarmiesse Let's do the pre-commit docker. Thank you!

Yes, all the public APIs are in __init__.py since we don't have much.
It might change in the future if the API is becoming heavy.

gabrieldemarmiesse:
Thanks @haifeng-jin . If we start having other public modules, we just need to add them to the list of modules to check :)

Issue #949, opened 2020-02-09
Finish the TODO list of classes/functions to type.


gabrieldemarmiesse:
We have a [big todo list of classes/functions to type](https://github.com/keras-team/autokeras/blob/dda2d9de0c602a1f47b45bb26783eacfd9572815/tests/autokeras/typed_api_test.py#L11) (add type hints, see [PEP484](https://www.python.org/dev/peps/pep-0484/) for more infos about type hints) and the docs: https://docs.python.org/3/library/typing.html

Functions of the list should be typed and for classes, only the contructor (`__init__()`) should be typed. No need to specify the return type of a `__init__()` because it's always `None`.

When you add types for a function/class, you can remove it from the TODO list in the same pull request.

Please make multiple small pull requests. Not a single big one :)

If you don't really understand what is asked:
We did the same procedure in tensorflow/addons here: https://github.com/tensorflow/addons/issues/989 See this issue and look into the pull requests made.

You can always ask if you're wondering what to do. It's fairly easy and a good way to start contributing to autokeras!

PR #957, opened 2020-02-11
Typed automodel and type checking during tests.

gabrieldemarmiesse:
Linked to #949


gabrieldemarmiesse:
Do not merge, I need to add a test to ensure the types are correct.

gabrieldemarmiesse:
We need to merge #960 before for this PR to work.

gabrieldemarmiesse:
About how this work: Typeguard has a pytest plugin. This plugin will hook into all autokeras imports and run the `typechecked` decorator. 

Meaning that during the tests, every time a function with type hints is called, the arguments are checked (as opposed to the default behavior in python, where type hints do nothing).

gabrieldemarmiesse:
Ready for review

autokeras/auto_model.py
```diff
@@ -87,15 +96,15 @@ class AutoModel(object):
...
-                 inputs,
-                 outputs,
-                 name='auto_model',
-                 max_trials=100,
-                 directory=None,
-                 objective='val_loss',
-                 tuner='greedy',
-                 overwrite=False,
-                 seed=None):
+                 inputs: Union[Node, List[Node]],
```
haifeng-jin: Please change Node to Input (ak.nodes.Input)

autokeras/auto_model.py
```diff
@@ -87,15 +96,15 @@ class AutoModel(object):
...
+                 outputs: Union[Block, Node, list],
```
haifeng-jin: Change Block to Head (ak.engine.head.Head)

setup.py
```diff
@@ -32,7 +32,8 @@
                   'isort',
                   'pytest-xdist',
                   'pytest-cov',
-                  'coverage'
+                  'coverage',
+                  'typeguard==2.7.1'
```
haifeng-jin: why are we pinning the version of this one?
Is it stable?

gabrieldemarmiesse: It's stable, I'm just a bit too careful sometimes haha

autokeras/auto_model.py
```diff
@@ -87,15 +95,15 @@ class AutoModel(object):
...
-                 outputs,
-                 name='auto_model',
-                 max_trials=100,
-                 directory=None,
-                 objective='val_loss',
-                 tuner='greedy',
-                 overwrite=False,
-                 seed=None):
+                 inputs: Union[Input, List[Input]],
+                 outputs: Union[head_module.Head, list],
```
haifeng-jin: It should be `Union[Head, Node, list]`. I think you accidentally removed `Node`.
Another concern is how will it show on the docs by autogen.py?
Will it be `head_module.Head` directly? That would be confusing for the reader.

gabrieldemarmiesse: Thanks, I'll fix it.

Good question, I'll dig into it. I'm planning to change the code generation too to display the types in the description, not the signature as it's too cumbersome and hurts readability.

gabrieldemarmiesse: Here is what it looks like in the docs: 

```
autokeras.AutoModel(
    inputs: Union[autokeras.nodes.Input, List[autokeras.nodes.Input]],
    outputs: Union[autokeras.engine.head.Head, list],
    name: str = "auto_model",
    max_trials: int = 100,
    directory: Union[str, pathlib.Path, None] = None,
    objective: str = "val_loss",
    tuner: Union[str, Type[autokeras.engine.tuner.AutoTuner]] = "greedy",
    overwrite: bool = False,
    seed: Optional[int] = None,
)
```

I can see two things to fix:

* Fist, use the aliases in the type hints instead of the complete class path.
* Second, display those info in the arguments descriptions only.

I'll open two issue in keras-autodoc. I'll fix the aliases issue first, and I believe we can merge it once it's done to avoid blocking the typing of autokeras.

The second issue is more or less making the docs beautiful, which is important, but not critical.

gabrieldemarmiesse: It looks like that now: 

```python
autokeras.AutoModel(
    inputs: Union[autokeras.Input, List[autokeras.Input]],
    outputs: Union[autokeras.Head, autokeras.Node, list],
    name: str = "auto_model",
    max_trials: int = 100,
    directory: Union[str, pathlib.Path, None] = None,
    objective: str = "val_loss",
    tuner: Union[str, Type[autokeras.engine.tuner.AutoTuner]] = "greedy",
    overwrite: bool = False,
    seed: Optional[int] = None,
)
```

`AutoTuner` is not listed in the docs, so there is no aliases, this is why keras-autodoc can't use an alias for it.

Do we plan to make `AutoTuner` public in the future?

haifeng-jin: Yes, we should make it public for anyone would like to override to implement their own tuner. I will merge this PR now.

Would you also help us make typedapi part of the setup.py tests dependencies? That would make the community easier to setup the environment. Thank you!

gabrieldemarmiesse: Sure thing, I'll upload it on pypi when I have the time :)

PR #978, opened 2020-02-17
Typed TextClassifier

gabrieldemarmiesse:
Related to #949

### DIFF
--- a/autokeras/tasks/text.py
+++ b/autokeras/tasks/text.py
@@ -2,6 +2,7 @@ import pathlib
 from typing import Callable
 from typing import List
 from typing import Optional
+from typing import Union
 
 from autokeras import auto_model
 from autokeras import hypermodels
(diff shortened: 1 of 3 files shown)


autokeras/tasks/text.py
```diff
@@ -36,16 +42,16 @@ class TextClassifier(SupervisedTextPipeline):
...
-                 name='text_classifier',
-                 max_trials=100,
-                 directory=None,
-                 objective='val_loss',
-                 overwrite=True,
-                 seed=None):
+                 num_classes: Optional[int] = None,
+                 multi_label: bool = False,
+                 loss: Callable = None,
+                 metrics: Optional[List[Callable]] = None,
```
haifeng-jin: I think loss and metrics should be more specific.
The callable should be keras loss and keras metric.
Loss can also be a string.
Metric can also be a list of strings.
They can also be dictionaries.
https://keras.io/models/model/#compile

gabrieldemarmiesse: Nice! This will clarify it for people reading the docs :)

haifeng-jin:
Screen Shot 2020-02-20 at 8 52 54 PM
@gabrieldemarmiesse 
It seems the docs doesn't display the keras types.

gabrieldemarmiesse:
Well, this is actually expected. `Union[Callable, Loss] ` == `Callable` since Loss is a callable. I just added the two since I believe it'd nicer to read in the code. I don't think there is anything we can do for the docs.

PR #980, opened 2020-02-17
Typing the XceptionBlock.

gabrieldemarmiesse:
Related to #949

A bit of background on this, `Literal` is quite useful for IDEs/type checkers to be able to identify arguments which can only take specific values. It works with `str`, `int`, `None` and `bool`. See here: https://docs.python.org/3/library/typing.html#typing.Literal 

But the problem is that it's available only for py3.8+. 

A backport was made, this is what we're using right now. See the section in the PEP: https://www.python.org/dev/peps/pep-0586/#backporting-the-literal-type

### DIFF
--- a/autokeras/hypermodels/basic.py
+++ b/autokeras/hypermodels/basic.py
@@ -1,3 +1,5 @@
+from typing import Optional
+
 from kerastuner.applications import resnet
 from kerastuner.applications import xception
 from tensorflow.keras import layers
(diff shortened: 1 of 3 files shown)


gabrieldemarmiesse:
Actually the docs isn't rendered well with this. Let's put that on hold until we can build the docs with python 3.8 (next tf release I think).

PR #983, opened 2020-02-18
Clarifying the supported python versions.

gabrieldemarmiesse:
The list of supported python versions wasn't very clear. I think we're compatible with python 3.6+, as such, we should test with the oldest version we support since python has strong backward compatibility guarantees.

The classifiers appear on pypi to make the search easier.


gabrieldemarmiesse:
I can't change the python version yet in github because typedapi doesn't support python 3.5. For the moment let's test with python 3.6, we should be able to catch more bugs than in python 3.7.

PR #990, opened 2020-02-19
Tao Chen First 

ECEMACHINE:

resolves #000


haifeng-jin:
@gabrieldemarmiesse I am thinking about use python notebooks like this and convert them into markdown in autogen.py to put the website.

I would like to use python notebooks for all the tutorials.
So that people can easily run them on Colab.

On the webpage of each tutorial, there should be a button like in this one to direct people to open in colab.
Do you think we should implement this feature in keras-autodoc or just do it in autokeras autogen.py?

gabrieldemarmiesse:
There is a project dedicated to this: https://pypi.org/project/mknotebooks/

For the moment, I believe it's better to use it directly in the autogen.py. If we see that there is a lot of boilerplate that we need to take care of, we'll integrate it into keras-autodoc with a nicer API.

haifeng-jin:
This one looks great. Thank you!

PR #1002, opened 2020-02-23
Typed DenseBlock

### LINKED ISSUE #949: Finish the TODO list of classes/functions to type.

sibyjackgrove:

related to #949 

Added types to DenseBlock init method.

### DIFF
--- a/autokeras/hypermodels/basic.py
+++ b/autokeras/hypermodels/basic.py
@@ -28,9 +28,9 @@ class DenseBlock(block_module.Block):
     """
 
     def __init__(self,
-                 num_layers=None,
-                 use_batchnorm=None,
-                 dropout_rate=None,
+                 num_layers: int = None,
+                 use_batchnorm: bool = None,
+                 dropout_rate: float = None,
                  **kwargs):
         super().__init__(**kwargs)
         self.num_layers = num_layers
--- a/autokeras/hypermodels/basic.py
+++ b/autokeras/hypermodels/basic.py
@@ -28,9 +28,9 @@ class DenseBlock(block_module.Block):
     """
 
     def __init__(self,
-                 num_layers: int = None,
-                 use_batchnorm: bool = None,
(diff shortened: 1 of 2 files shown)


autokeras/hypermodels/basic.py
```diff
@@ -28,9 +28,9 @@ class DenseBlock(block_module.Block):
     """
 
     def __init__(self,
-                 num_layers=None,
-                 use_batchnorm=None,
-                 dropout_rate=None,
+                 num_layers: int = None,
+                 use_batchnorm: bool = None,
+                 dropout_rate: float = None,
```
haifeng-jin: These three args should all be optional. You may refer to other typed blocks.

gabrieldemarmiesse:
Be careful when writing "Resolves #949", it's going to close the issue when merged.

autokeras/hypermodels/basic.py
sibyjackgrove: Changed to optional with 55d0c99.

PR #1014, opened 2020-02-28
1st version Image Segmenter

ECEMACHINE:

resolves #000

### DIFF
--- a/autokeras/nodes.py
+++ b/autokeras/nodes.py
@@ -25,7 +25,8 @@ class ImageInput(Input):
     """Input node for image data.
 
     The input data should be numpy.ndarray or tf.data.Dataset. The shape of the data
-    should be 3 or 4 dimensional, the last dimension of which should be channel
+    should be 3 or 4 dimensional, the first three dimensions should be samples, width
+    and height of the image dataset,the last dimension of which should be channel
     dimension.
     """
 
--- a/autokeras/nodes.py
+++ b/autokeras/nodes.py
@@ -25,9 +25,8 @@ class ImageInput(Input):
     """Input node for image data.
 
     The input data should be numpy.ndarray or tf.data.Dataset. The shape of the data
-    should be 3 or 4 dimensional, the first three dimensions should be samples, width
-    and height of the image dataset,the last dimension of which should be channel
-    dimension.
+    should be should be (samples, width, height) or
+    (samples, width, height, channels).
     """
 
     def get_adapter(self):
--- a/autokeras/tasks/image.py
+++ b/autokeras/tasks/image.py
@@ -73,7 +73,7 @@ class ImageClassifier(SupervisedImagePipeline):
             y: Any = None,
             epochs: Optional[int] = None,
             callbacks: Optional[List[Union[str, Callable]]] = None,
-            validation_split: Optional[int] = 0.2,
+            validation_split=0.2,
             validation_data: Any = None,
             **kwargs):
         """Search for the best model and hyperparameters for the AutoModel.
@@ -173,7 +173,7 @@ class ImageRegressor(SupervisedImagePipeline):
             y: Any = None,
             epochs: Optional[int] = None,
             callbacks: Optional[List[Union[str, Callable]]] = None,
-            validation_split: Optional[int] = 0.2,
+            validation_split=0.2,
             validation_data: Any = None,
             **kwargs):
         """Search for the best model and hyperparameters for the AutoModel.
@@ -279,7 +279,7 @@ class ImageSegmenter(SupervisedImagePipeline):
             y: Any = None,
             epochs: Optional[int] = None,
             callbacks: Optional[List[Union[str, Callable]]] = None,
-            validation_split: Optional[int] = 0.2,
+            validation_split=0.2,
             validation_data: Any = None,
             **kwargs):
         """Search for the best model and hyperparameters for the AutoModel.
--- a/autokeras/tasks/image.py
+++ b/autokeras/tasks/image.py
@@ -23,7 +23,7 @@ class ImageClassifier(SupervisedImagePipeline):
     """AutoKeras image classification class.
 
     # Arguments
-        num_classes: Int. Defaults to None. If None, it will infer from the data.
+        num_classes: Int. Defaults to None. If None, it will be inferred from the data.
         multi_label: Boolean. Defaults to False.
         loss: A Keras loss function. Defaults to use 'binary_crossentropy' or
             'categorical_crossentropy' based on the number of classes.
@@ -81,8 +81,9 @@ class ImageClassifier(SupervisedImagePipeline):
 
         # Arguments
             x: numpy.ndarray or tensorflow.Dataset. Training data x. The shape of the
-                data should be 3 or 4 dimensional, the last dimension of which should
-                be channel dimension.
+                data should be 3 or 4 dimensional, the first three dimensions should be
+                samples, width and height of the image dataset,the last dimension of
+                which should be channel dimension.
             y: numpy.ndarray or tensorflow.Dataset. Training data y. It can be raw
                 labels, one-hot encoded if more than two classes, or binary encoded
                 for binary classification.
@@ -127,7 +128,7 @@ class ImageRegressor(SupervisedImagePipeline):
 
     # Arguments
         output_dim: Int. The number of output dimensions. Defaults to None.
-            If None, it will infer from the data.
+            If None, it will be inferred from the data.
         loss: A Keras loss function. Defaults to use 'mean_squared_error'.
         metrics: A list of Keras metrics. Defaults to use 'mean_squared_error'.
         name: String. The name of the AutoModel. Defaults to 'image_regressor'.
@@ -145,15 +146,15 @@ class ImageRegressor(SupervisedImagePipeline):
     """
 
     def __init__(self,
-                 output_dim=None,
-                 loss='mean_squared_error',
-                 metrics=None,
-                 name='image_regressor',
-                 max_trials=100,
-                 directory=None,
-                 objective='val_loss',
-                 overwrite=True,
-                 seed=None):
+                 output_dim : Optional[int] = None,
+                 loss: str = 'mean_squared_error',
+                 metrics: Optional[List[Union[str, Callable]]] = None,
+                 name: str = 'image_regressor',
+                 max_trials: int = 100,
+                 directory: Union[str, Path, None] = None,
+                 objective: str = 'val_loss',
+                 overwrite: bool = True,
+                 seed: Optional[int] = None):
         super().__init__(
             outputs=hypermodels.RegressionHead(output_dim=output_dim,
                                                loss=loss,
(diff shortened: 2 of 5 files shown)


autokeras/tasks/image.py
```diff
@@ -221,3 +221,103 @@ def fit(self,
...
+
+
+class ImageSegmenter(SupervisedImagePipeline):
+    """AutoKeras image segmentation class.
+
+    # Arguments
+        num_classes: Int. Defaults to None. If None, it will infer from the data.
+        loss: A loss function. Defaults to use 'tf.losses.softmax_cross_entropy'.
+        metrics: A list of metrics used to measure the accuracy of the model,
+        default to 'mIoU'
```
haifeng-jin: 1. Add an indent here.
2. a period in the end.
3. Let's use accuracy as default value.

autokeras/tasks/image.py
```diff
@@ -221,3 +221,103 @@ def fit(self,
...
                     validation_data=validation_data,
                     **kwargs)
+
+
+class ImageSegmenter(SupervisedImagePipeline):
+    """AutoKeras image segmentation class.
+
+    # Arguments
+        num_classes: Int. Defaults to None. If None, it will infer from the data.
+        loss: A loss function. Defaults to use 'tf.losses.softmax_cross_entropy'.
```
haifeng-jin: We should use the same description as the image classifier.

autokeras/tasks/image.py
```diff
@@ -221,3 +221,104 @@ def fit(self,
...
+
+        It will search for the best model based on the performances on
+        validation data.
+
+        # Arguments
+            x: numpy.ndarray or tensorflow.Dataset. Training image dataset x.
+                The shape of the data should be 3 or 4 dimensional, the last
+                dimension of which should be channel dimension.
+            y: numpy.ndarray or tensorflow.Dataset. Training image data
+                set y. It should be the groundtruth image data set as the label.
```
haifeng-jin: Change "It should be the groundtruth image data set as the label." to "It should be a tensor with the same shape of x. Each element in the tensor is the label of the corresponding pixel."

haifeng-jin: https://github.com/keras-team/autokeras/compare/33879915c9a2ef9c1348170149c35dc13773fd41..ad552e780c02799cb3b64059ba08a278ce55e5f3#diff-173d8f18374c208cb0df47eb0bc0e6dbL266

haifeng-jin: https://github.com/keras-team/autokeras/compare/33879915c9a2ef9c1348170149c35dc13773fd41..ad552e780c02799cb3b64059ba08a278ce55e5f3#diff-173d8f18374c208cb0df47eb0bc0e6dbL288-L289

haifeng-jin: https://github.com/keras-team/autokeras/compare/33879915c9a2ef9c1348170149c35dc13773fd41..ad552e780c02799cb3b64059ba08a278ce55e5f3#diff-173d8f18374c208cb0df47eb0bc0e6dbL227

haifeng-jin: These lines need to be changed back to it was before the force push. Try not to force-push when multiple people have committed to the same branch.

autokeras/tasks/image.py
```diff
@@ -221,3 +221,107 @@ def fit(self,
...
+        super().__init__(
+            outputs=hypermodels.SegmenterHead(num_classes=num_classes,
+                                              multi_label=multi_label,
+                                              loss=loss,
+                                              metrics=metrics),
+            max_trials=max_trials,
+            directory=directory,
+            name=name,
+            objective=objective,
+            tuner=task_specific.ImageSegmenterTuner,
```
haifeng-jin: Let's use `greedy.Greedy` for now.

autokeras/tasks/image.py
```diff
@@ -221,3 +221,107 @@ def fit(self,
                     validation_split=validation_split,
                     validation_data=validation_data,
                     **kwargs)
+
+
+class ImageSegmenter(SupervisedImagePipeline):
+    """AutoKeras image segmentation class.
+
+    # Arguments
+        num_classes: Int. Defaults to None. If None, it will infer from the data.
```
fchollet: "it will be inferred from" (you can also apply this fix to the other docstrings)

autokeras/tasks/image.py
```diff
@@ -221,3 +221,107 @@ def fit(self,
                     validation_split=validation_split,
                     validation_data=validation_data,
                     **kwargs)
+
+
+class ImageSegmenter(SupervisedImagePipeline):
+    """AutoKeras image segmentation class.
+
+    # Arguments
```
fchollet: Consider including type annotations.

autokeras/tasks/image.py
```diff
@@ -221,3 +221,107 @@ def fit(self,
...
+            **kwargs):
+        """Search for the best model and hyperparameters for the AutoModel.
+
+        It will search for the best model based on the performances on
+        validation data.
+
+        # Arguments
+            x: numpy.ndarray or tensorflow.Dataset. Training image dataset x.
+                The shape of the data should be 3 or 4 dimensional, the last
+                dimension of which should be channel dimension.
```
fchollet: This is rather unclear. Please provide clarification on what the expect input format is. Intuitively one would expect `(samples, width, height, channels)`, so it's not clear where the 3D comes from.

autokeras/tasks/image.py
```diff
@@ -221,3 +221,107 @@ def fit(self,
...
+            y: numpy.ndarray or tensorflow.Dataset. Training image data
+                set y. It should be a tensor with the same shape of x. Each
```
fchollet: It's not the same shape since the number of channels will be different. It's the same *height* and *width*.

haifeng-jin:
@ECEMACHINE Would you please help address Francois's comments?
Note some docstring problems may also exist in the docstrings of other classes.
Please check ImageClassifier, ImageRegressor, ImageInput for similar problems.

Thank you!

autokeras/nodes.py
```diff
@@ -25,7 +25,8 @@ class ImageInput(Input):
     """Input node for image data.
 
     The input data should be numpy.ndarray or tf.data.Dataset. The shape of the data
-    should be 3 or 4 dimensional, the last dimension of which should be channel
+    should be 3 or 4 dimensional, the first three dimensions should be samples, width
+    and height of the image dataset,the last dimension of which should be channel
     dimension.
```
haifeng-jin: Change it to "should be (samples, width, height) or (samples, width, height, channels)."
Please also follow this style to change the rest.

autokeras/tasks/image.py
```diff
@@ -23,7 +23,8 @@ class ImageClassifier(SupervisedImagePipeline):
     """AutoKeras image classification class.
 
     # Arguments
-        num_classes: Int. Defaults to None. If None, it will infer from the data.
+        num_classes: Int. Defaults to None. If None, it will be inferred from the
```
haifeng-jin: I just realized there may be more of this grammar error. Would you please use a global search to find them all and change them?

autokeras/tasks/image.py
```diff
@@ -221,3 +224,110 @@ def fit(self,
...
+            overwrite=overwrite,
+            seed=seed)
+
+    def fit(self,
+            x=None,
+            y=None,
+            epochs=None,
+            callbacks=None,
+            validation_split=0.2,
+            validation_data=None,
```
haifeng-jin: We need type annotations to these arguments as well.

autokeras/tasks/image.py
```diff
@@ -145,15 +147,15 @@ class ImageRegressor(SupervisedImagePipeline):
...
-                 metrics=None,
-                 name='image_regressor',
-                 max_trials=100,
-                 directory=None,
-                 objective='val_loss',
-                 overwrite=True,
-                 seed=None):
+                 output_dim: Optional[int] = None,
+                 loss: str = 'mean_squared_error',
+                 metrics: Optional[List[Union[str, Callable]]] = None,
```
haifeng-jin: Please follow
https://github.com/keras-team/autokeras/blob/7cade0d7853f25d766f35527df14cbd4fdfedb74/autokeras/tasks/text.py#L47
https://github.com/keras-team/autokeras/blob/7cade0d7853f25d766f35527df14cbd4fdfedb74/autokeras/tasks/text.py#L48
to add the type for loss and metrics.  And for the ImageClassifier and ImageSegmenter, too. The original typing for ImageClassifier may not be correct.

autokeras/tasks/image.py
```diff
@@ -167,22 +169,22 @@ def __init__(self,
...
 
     def fit(self,
-            x=None,
-            y=None,
-            epochs=None,
-            callbacks=None,
-            validation_split=0.2,
-            validation_data=None,
+            x: Any = None,
+            y: Any = None,
```
haifeng-jin: They should be `Union[np.ndarray, tf.data.Dataset]`.

autokeras/tasks/image.py
```diff
@@ -167,22 +169,22 @@ def __init__(self,
...
+            epochs: Optional[int] = None,
+            callbacks: Optional[List[Union[str, Callable]]] = None,
```
haifeng-jin: It should be `Optional[List[tf.keras.callbacks.Callback]]`.

autokeras/tasks/image.py
```diff
@@ -145,15 +150,15 @@ class ImageRegressor(SupervisedImagePipeline):
...
-                 loss='mean_squared_error',
-                 metrics=None,
-                 name='image_regressor',
-                 max_trials=100,
-                 directory=None,
-                 objective='val_loss',
-                 overwrite=True,
-                 seed=None):
+                 output_dim: Optional[int] = None,
+                 loss: utils.AcceptableLoss = None,
```
haifeng-jin: The default value should be `mean_squared_error`.

autokeras/tasks/image.py
```diff
@@ -67,22 +72,22 @@ def __init__(self,
             seed=seed)
 
     def fit(self,
-            x=None,
-            y=None,
-            epochs=None,
-            callbacks=None,
-            validation_split=0.2,
-            validation_data=None,
+            x: Union[np.ndarray, tf.data.Dataset] = None,
```
haifeng-jin: Change to `x: Optional[Union[np.ndarray, tf.data.Dataset]] = None,`

autokeras/tasks/image.py
```diff
@@ -67,22 +72,22 @@ def __init__(self,
...
+            y: Union[np.ndarray, tf.data.Dataset] = None,
```
haifeng-jin: This one should be optional, too.

autokeras/tasks/image.py
```diff
@@ -221,3 +226,109 @@ def fit(self,
...
+                 num_classes: Optional[int] = None,
+                 multi_label: bool = False,
+                 loss: utils.AcceptableLoss = None,
+                 metrics: utils.AcceptableMetrics = None,
+                 name: str = 'image_classifier',
+                 max_trials: int = 100,
+                 directory: Union[str, Path, None] = None,
+                 objective: str = 'val_loss',
+                 overwrite: bool = True,
+                 seed: Optional[int] = None,):
```
haifeng-jin: remove the `,`

autokeras/tasks/image.py
```diff
@@ -67,22 +72,22 @@ def __init__(self,
...
-            epochs=None,
-            callbacks=None,
-            validation_split=0.2,
-            validation_data=None,
+            x: Union[np.ndarray, tf.data.Dataset] = None,
+            y: Union[np.ndarray, tf.data.Dataset] = None,
+            epochs: Optional[int] = None,
+            callbacks: Optional[List[tf.keras.callbacks.Callback]] = None,
+            validation_split: Optional[float] = 0.2,
+            validation_data: Any = None,
```
haifeng-jin: Change to `validation_data: Optional[Union[np.ndarray, tf.data.Dataset]] = None,`

PR #1017, opened 2020-03-02
Time Series Input Adapters

### LINKED ISSUE #782: TimeSeriesForecaster

abgese:

resolves #782

PR to handle inputs for time series data

### DIFF
--- a/autokeras/adapters/input_adapter.py
+++ b/autokeras/adapters/input_adapter.py
@@ -193,7 +193,7 @@ class StructuredDataInputAdapter(adapter_module.Adapter):
                 self.column_types[key] = value
 
 
-class TimeSeriesInputAdapter(adapter_module.Adapter):
+class TimeseriesInputAdapter(adapter_module.Adapter):
 
     def __init__(self,
                  lookback=None,
@@ -209,7 +209,7 @@ class TimeSeriesInputAdapter(adapter_module.Adapter):
         config = super().get_config()
         config.update({
             'lookback': self.lookback,
-            'column_name': self.column_names,
+            'column_names': self.column_names,
             'column_types': self.column_types
         })
         return config
--- a/autokeras/adapters/input_adapter.py
+++ b/autokeras/adapters/input_adapter.py
@@ -191,3 +191,49 @@ class StructuredDataInputAdapter(adapter_module.Adapter):
         for key, value in column_types.items():
             if key not in self.column_types:
                 self.column_types[key] = value
+
+
+class TimeSeriesInputAdapter(adapter_module.Adapter):
+
+    def __init__(self, lookback=None, **kwargs):
+        super().__init__(**kwargs)
+        self.lookback = lookback
+
+    def get_config(self):
+        config = super().get_config()
+        config.update({
+            'lookback': self.lookback,
+        })
+        return config
+
+    @classmethod
+    def from_config(cls, config):
+        obj = super().from_config(config)
+        obj.lookback = config['lookback']
+
+    def check(self, x):
+        """Record any information needed by transform."""
+        if not isinstance(x, (np.ndarray, tf.data.Dataset)):
+            raise TypeError('Expect the data in TimeSeriesInput to be numpy.ndarray'
+                            ' or tf.data.Dataset, but got {type}.'.format(
+                                type=type(x)))
+
+        if isinstance(x, np.ndarray) and x.ndim != 2:
+            raise ValueError('Expect the data in TimeSeriesInput to have 2 dimension'
+                             ', but got input shape {shape} with {ndim} '
+                             'dimensions'.format(
+                                 shape=x.shape,
+                                 ndim=x.ndim))
+        if isinstance(x, np.ndarray) and not np.issubdtype(x.dtype, np.number):
+            raise TypeError('Expect the data in TimeSeriesInput to be numerical, but'
+                            'got {type}.'.format(type=x.dtype))
+
+    def convert_to_dataset(self, x):
+        if isinstance(x, np.ndarray):
+            x = tf.data.Dataset.from_tensor_slices(x)
+        x = x.window(self.lookback, shift=1, drop_remainder=True)
+        final_data = []
+        for window in x:
+            final_data.append([elems.numpy() for elems in window])
+        final_data = tf.data.Dataset.from_tensor_slices(final_data)
+        return final_data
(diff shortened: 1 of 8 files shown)


autokeras/adapters/input_adapter.py
```diff
@@ -191,3 +191,49 @@ def infer_column_types(self):
...
+                                type=type(x)))
+
+        if isinstance(x, np.ndarray) and x.ndim != 2:
+            raise ValueError('Expect the data in TimeSeriesInput to have 2 dimension'
+                             ', but got input shape {shape} with {ndim} '
+                             'dimensions'.format(
+                                 shape=x.shape,
+                                 ndim=x.ndim))
+        if isinstance(x, np.ndarray) and not np.issubdtype(x.dtype, np.number):
+            raise TypeError('Expect the data in TimeSeriesInput to be numerical, but'
```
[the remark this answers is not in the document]
abgese: I have a doubt, If we support categorical features, then wont we have to keep the dtype of the transformed output in the convert_to_dataset function to be unicode (Like StructuredDataInputAdapter). Will the later Blocks in autokeras be able to handle that?

autokeras/adapters/input_adapter.py
```diff
@@ -191,3 +191,78 @@ def infer_column_types(self):
...
+            if self.column_types:
+                raise ValueError('Column names must be specified.')
+            self.column_names = [index for index in range(x.shape[1])]
+
+        # Check if column_names has the correct length.
+        if len(self.column_names) != x.shape[1]:
+            raise ValueError('Expect column_names to have length {expect} '
+                             'but got {actual}.'.format(
+                                 expect=x.shape[1],
+                                 actual=len(self.column_names)))
```
haifeng-jin: Can we extract this part as a separate member function and let TimeSeriesInputAdapter to subclass StructuredDataInputAdapter? or we can extract a superclass for both of them to subclass.

autokeras/nodes.py
```diff
@@ -96,4 +96,42 @@ def config_from_adapter(self, adapter):
...
 class TimeSeriesInput(Input):
-    pass
+    """Input node for Time Series data.
+
+    # Arguments
+        lookback: Int. The range of history steps to consider for each prediction.
+            For example, if lookback=n, the data in the range of [i - n, i - 1]
+            is used to predict the value of step i. If unspecified, it will be tuned
+            automatically.
+    """
```
haifeng-jin: We need the docstrings for column_names and column_types here.

haifeng-jin:
@abgese It seems the tests are not covering the TimeSeriesInput. Would you please add a test to cover it? Thanks.

autokeras/adapters/input_adapter.py
abgese: One issue is that the "update" function in the "StructuredDataInputAdapter" expects every row to be a vector of single values. But after the "convert_to_dataset" function ini "TimeSeriesInputAdapter" the dataset is transformed with every row to be a vector of vectors. This causes the "decode" function to fail. 

Also, the concept of column_names loses it's meaning after the data transform , so is it really relevant information that needs to be kept in the class?

autokeras/nodes.py
abgese: Added, but do we really need column_names and column_types? because after teh data transform the "columns" in the transformed dataset is a vector instead of the original single value.

autokeras/adapters/input_adapter.py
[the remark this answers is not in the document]
haifeng-jin: Yes, we can handle it. After tf 2.2 release, we will use the official categorical encoding layers to handle it.

autokeras/nodes.py
haifeng-jin: Yes, every arg should be accessible even after use.

autokeras/adapters/input_adapter.py
haifeng-jin: You can use whatever way you think makes the most sense to reuse the this part of check code. Superclass is one way, or you can use a stand alone-function.

Yes, we need to keep it. The user may want to access it later.

abgese: Alright for now I've used reused the check function. Since that is the only thing we really need.

autokeras/adapters/input_adapter.py
```diff
@@ -191,3 +191,78 @@ def infer_column_types(self):
...
+        super().__init__(**kwargs)
+        self.lookback = lookback
+        self.column_names = column_names
+        self.column_types = column_types
+
+    def get_config(self):
+        config = super().get_config()
+        config.update({
+            'lookback': self.lookback,
+            'column_name': self.column_names,
```
fchollet: `column_names` (missing s). Please make sure you have appropriate tests in order to catch such issues.

autokeras/nodes.py
```diff
@@ -96,4 +96,54 @@ def config_from_adapter(self, adapter):
 
 
 class TimeSeriesInput(Input):
-    pass
+    """Input node for Time Series data.
```
fchollet: timeseries

autokeras/nodes.py
```diff
@@ -96,4 +96,54 @@ def config_from_adapter(self, adapter):
 
 
 class TimeSeriesInput(Input):
```
fchollet: TimeseriesInput (no need to capitalize series)

autokeras/adapters/input_adapter.py
```diff
@@ -191,3 +191,78 @@ def infer_column_types(self):
         for key, value in column_types.items():
             if key not in self.column_types:
                 self.column_types[key] = value
+
+
+class TimeSeriesInputAdapter(adapter_module.Adapter):
```
fchollet: TimeseriesInputAdapter

autokeras/nodes.py
```diff
@@ -96,4 +96,54 @@ def config_from_adapter(self, adapter):
...
+        self.column_types = column_types
+
+    def build(self):
+        return tf.keras.Input(shape=self.shape, dtype=tf.float32)
+
+    def get_config(self):
+        config = super().get_config()
+        config.update({
+            'lookback': self.lookback,
+            'column_name': self.column_names,
```
fchollet: Same error here

autokeras/adapters/input_adapter.py
haifeng-jin: @abgese Would you please add a file `tests/node_test.py` to use these two functions to test the config functions of the nodes? https://github.com/keras-team/autokeras/blob/master/autokeras/graph.py#L35-L45 You may find how to use these two functions in other parts of the code. Thank you!

abgese: Done

PR #1056, opened 2020-03-21
Image segmenter

ECEMACHINE:

resolves #000

### DIFF
--- a/autokeras/hypermodels/heads.py
+++ b/autokeras/hypermodels/heads.py
@@ -23,7 +23,8 @@ class ClassificationHead(head_module.Head):
     or one-hot encoded if more than two classes were found.
 
     # Arguments
-        num_classes: Int. Defaults to None. If None, it will infer from the data.
+        num_classes: Int. Defaults to None. If None, it will be inferred from the
+            data.
         multi_label: Boolean. Defaults to False.
         loss: A Keras loss function. Defaults to use `binary_crossentropy` or
             `categorical_crossentropy` based on the number of classes.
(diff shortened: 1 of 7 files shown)


autokeras/hypermodels/heads.py
```diff
@@ -169,3 +170,102 @@ def build(self, hp, inputs=None):
...
+    Use sigmoid and binary crossentropy for binary classification and multi-label
+    classification. Use softmax and categorical crossentropy for multi-class
+    (more than 2) segmentation. Use Accuracy as metrics by default.
+
+    The targets passing to the head would have to be tf.data.Dataset, np.ndarray,
+    pd.DataFrame or pd.Series. It can be raw labels, one-hot encoded if more than two
+    classes, or binary encoded for binary classification.
+
+    The raw labels will be encoded to one column if two classes were found,
+    or one-hot encoded if more than two classes were found.
```
haifeng-jin: This description needs to be changed to the segmentation setting. 
"The raw labels will be encoded to 0s and 1s if two classes were found, or one-hot encoded if more than two classes were found."

autokeras/hypermodels/heads.py
```diff
@@ -169,3 +170,102 @@ def build(self, hp, inputs=None):
...
+
+    # Arguments
+        num_classes: Int. Defaults to None. If None, it will be inferred from the
+            data.
+        multi_label: Boolean. Defaults to False.
```
haifeng-jin: We don't support multi_label for the segmentation for now. Please remove this argument

autokeras/hypermodels/heads.py
```diff
@@ -169,3 +170,102 @@ def build(self, hp, inputs=None):
 
     def get_adapter(self):
         return adapters.RegressionHeadAdapter(name=self.name)
+
+
+class SegmenterHead(head_module.Head):
```
haifeng-jin: I think we should extend the ClassificationHead class. They share a lot in common.

PR #1072, opened 2020-03-29
Implement the Image Augmentation block with preprocessing layers.

haifeng-jin:

Implement the Image Augmentation block with preprocessing layers.

### DIFF
--- a/autokeras/hypermodels/preprocessing.py
+++ b/autokeras/hypermodels/preprocessing.py
@@ -140,13 +140,13 @@ class ImageAugmentation(block_module.Block):
         self.shape = None
 
     def build(self, hp, inputs=None):
-        list(
+        list([
             preprocessing.RandomContrast,
             preprocessing.RandomTranslation,
             preprocessing.RandomFlip,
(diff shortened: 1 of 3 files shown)


autokeras/hypermodels/preprocessing.py
```diff
@@ -121,26 +122,71 @@ class ImageAugmentation(block_module.Block):
     """
 
     def __init__(self,
-                 translation=True,
-                 horizontal_flip=True,
-                 vertical_flip=True,
+                 translation_range=0.5,
```
fchollet, line 125: About `translation_range` and `rotation_range`: they aren't quite ranges (a range implies two values, e.g. `[-0.5, 0.5]`). We call this argument `factor` in `RandomRotation` (and `height/width_factor` in RandomTranslation). `rotation_factor`, `translation_factor` would seem appropriate here. What do you think?

haifeng-jin: Yes, those better names. I will rename them.

PR #1083, opened 2020-04-04
ImageSegmentationHead

ECEMACHINE:

resolves #000


autokeras/hypermodels/heads.py
```diff
@@ -173,3 +173,66 @@ def build(self, hp, inputs=None):
...
+        super().__init__(loss=loss,
+                         metrics=metrics,
+                         num_classes=num_classes,
+                         dropout_rate=dropout_rate,
+                         **kwargs)
+        self.num_classes = num_classes
+        if not self.metrics:
+            self.metrics = ['accuracy']
+        self.dropout_rate = dropout_rate
+        self.set_loss()
```
haifeng-jin, line 218: We don't need these lines.
They are handled in the super class.

autokeras/hypermodels/heads.py
```diff
@@ -173,3 +173,66 @@ def build(self, hp, inputs=None):
...
+
+    def set_loss(self):
+        super().set_loss()
+
+    def get_config(self):
+        return super().get_config()
```
haifeng-jin, line 224: We don't need to override a method if it is only calling the super class method.
They will call the super class method by default.

autokeras/hypermodels/heads.py
```diff
@@ -173,3 +173,66 @@ def build(self, hp, inputs=None):
...
+
+    def build(self, hp, inputs=None):
```
haifeng-jin, line 226: The build function should be different from image classification.
You can refer to an example of image segmentation implemented using Keras.
See how they implemented the head part.

autokeras/hypermodels/heads.py
```diff
@@ -173,3 +173,66 @@ def build(self, hp, inputs=None):
...
+        return super().build(
+            hp=hp,
+            inputs=inputs
+        )
+
+    def get_adapter(self):
+        return adapters.ClassificationHeadAdapter(name=self.name)
```
haifeng-jin, line 233: We need a class of SegmentationHeadAdapter, which extends ClassificationHeadAdapter, in `output_adpater.py`.
It can be empty for now.
Just use `pass` as the content of the class.

autokeras/hypermodels/heads.py
```diff
@@ -173,3 +173,66 @@ def build(self, hp, inputs=None):
...
+
+    def config_from_adapter(self, adapter):
```
haifeng-jin, line 235: The same here. If we don't have anything different, we don't override.

PR #1134, opened 2020-05-17
add ngram to TextToNgramVectors

haifeng-jin:

resolves #000

Added Ngram in the search space.
As adding tests to the block, refactored the block tests.

### DIFF
--- a/autokeras/hypermodels/preprocessing.py
+++ b/autokeras/hypermodels/preprocessing.py
@@ -1,4 +1,6 @@
 from typing import Optional
+from typing import Tuple
+from typing import Union
 
 from tensorflow.keras.layers.experimental import preprocessing
 from tensorflow.python.util import nest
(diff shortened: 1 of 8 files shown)


autokeras/hypermodels/preprocessing.py
```diff
@@ -76,23 +78,37 @@ class TextToNgramVector(block_module.Block):
...
         super().__init__(**kwargs)
         self.max_tokens = max_tokens
+        self.ngrams = ngrams
 
     def build(self, hp, inputs=None):
         input_node = nest.flatten(inputs)[0]
+        if self.ngrams is not None:
+            ngrams = self.ngrams
+        else:
+            ngrams = hp.Int('ngrams', min_value=1, max_value=3)
```
fchollet, line 100: We may want to restrict `max_value` to 2 for large values of `max_tokens` (>20000). Indexing 3-grams would be too expensive over a very large corpus and vocabulary.

PR #1139, opened 2020-05-20
refactored input check for automodel

### LINKED ISSUE #1119: ImageClassifier with tf.data.Dataset input - classification head set loss error
### Bug Description

The problem may be caused by the ClassificationHeadAdapter- > fit_before_convert method, which does not convert the tf TensorShape to integer. Not sure if in newer version of tf, the comparison between tf TensorShape and integer is allowed.

### Bug Reproduction
codes are in the image classifier example in jupyter (data format section)

Data used by the code:
MNIST

### Expected Behavior

Screen Shot 2020-05-05 at 11 50 49 AM

### Setup Details
 - OS type and version:
 - Python: 3.6
 - autokeras: 1.0.2
 - keras-tuner:
 - scikit-learn:
 - numpy:
 - pandas:
 - tensorflow: 2.1.0

haifeng-jin:

resolves #1119 

The bug is caused by not correctly mapping the x and y to the right input and output node.
The code structure of this part is refactored and more checks are added.

### DIFF
--- a/autokeras/auto_model.py
+++ b/autokeras/auto_model.py
@@ -324,14 +324,14 @@ class AutoModel(object):
 
         if len(x_shapes) != len(self.inputs):
             raise ValueError(
-                'Expect x{in_val} to have {input_num} sources, '
+                'Expect x{in_val} to have {input_num} arrays, '
                 'but got {data_num}'.format(
                     in_val=in_val,
                     input_num=len(self.inputs),
                     data_num=len(x_shapes)))
         if len(y_shapes) != len(self.outputs):
             raise ValueError(
-                'Expect y{in_val} to have {output_num} sources, '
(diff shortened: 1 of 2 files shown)


autokeras/auto_model.py
```diff
@@ -288,16 +289,54 @@ def _process_xy(self, x, y, fit):
...
+        if len(x_shapes) != len(self.inputs):
+            raise ValueError(
+                'Expect x{in_val} to have {input_num} sources, '
+                'but got {data_num}'.format(
+                    in_val=in_val,
+                    input_num=len(self.inputs),
+                    data_num=len(x_shapes)))
+        if len(y_shapes) != len(self.outputs):
+            raise ValueError(
+                'Expect y{in_val} to have {output_num} sources, '
```
fchollet, line 334: sources -> arrays

PR #1142, opened 2020-05-20
Adapt Preprocessing layers before fit

haifeng-jin:

resolves #000

### DIFF
--- a/autokeras/engine/tuner.py
+++ b/autokeras/engine/tuner.py
@@ -49,11 +49,16 @@ class AutoTuner(kerastuner.engine.tuner.Tuner):
         return model
 
     def _on_train_end(self, model, hp, x, *args, **kwargs):
+        """Adapt the preprocessing layers and tune the fit arguments."""
         self.adapt(model, x)
 
     @staticmethod
     def adapt(model, dataset):
         """Adapt the preprocessing layers in the model."""
+        # Currently, only support using the original dataset to adapt all the
+        # preprocessing layers before the first non-preprocessing layer.
+        # TODO: Use PreprocessingStage for preprocessing layers adapt.
+        # TODO: Use Keras Tuner for preprocessing layers adapt.
         x = dataset.map(lambda x, y: x)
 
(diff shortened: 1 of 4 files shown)


autokeras/engine/tuner.py
```diff
@@ -47,46 +48,31 @@ def get_best_model(self):
...
-            **fit_kwargs)
-
-    def build_preprocessors(self, hp, x=None, validation_data=None):
-        # TODO: Split x and build the preprocessor for each input node.
-        return x, validation_data
+    def _on_train_end(self, model, hp, x, *args, **kwargs):
+        self.adapt(model, x)
+
+    @staticmethod
+    def adapt(model, dataset):
```
fchollet, line 55: The logic in this method will only work for single-input, single-output stacks of layers (Sequential or Sequential-like models). Will it be necessary to support arbitrary topologies in the future?

autokeras/engine/tuner.py
```diff
@@ -47,46 +48,31 @@ def get_best_model(self):
...
-            trial=trial,
-            x=x,
-            validation_data=validation_data,
-            *fit_args,
-            **fit_kwargs)
-
-    def build_preprocessors(self, hp, x=None, validation_data=None):
-        # TODO: Split x and build the preprocessor for each input node.
-        return x, validation_data
+    def _on_train_end(self, model, hp, x, *args, **kwargs):
```
fchollet, line 51: What is the purpose of this method? Please add a docstring.

autokeras/engine/tuner.py
haifeng-jin: Yes, the logic now is quite limited, but good enough to support the basic use cases of AutoKeras now. We will leverage [PreprocessingStage](https://github.com/keras-team/governance/blob/master/rfcs/20190502-preprocessing-layers.md#prepocessingstage) to improve it in the future. Eventually, this may be moved to keras tuner for supporting preprocessing layers.

PR #1144, opened 2020-05-24
fixing prediction with dataset for multi io model

haifeng-jin:

resolves #000

### DIFF
--- a/autokeras/auto_model.py
+++ b/autokeras/auto_model.py
@@ -270,40 +270,51 @@ class AutoModel(object):
         for source, hm, adapter in zip(sources, hms, adapters):
             if fit:
                 source = adapter.fit_transform(source)
+                hm.config_from_adapter(adapter)
             else:
                 source = adapter.transform(source)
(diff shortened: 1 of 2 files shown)


autokeras/auto_model.py
```diff
@@ -364,6 +378,23 @@ def _prepare_data(self, x, y, validation_data, validation_split):
                 validation_split)
         return dataset, validation_data
 
+    def _get_x(self, dataset):
```
fchollet, line 381: This method would benefit form a docstring, since the name is not self-explanatory

autokeras/auto_model.py
```diff
@@ -364,6 +378,23 @@ def _prepare_data(self, x, y, validation_data, validation_split):
...
+        shapes = data_utils.dataset_shape(dataset)
+        # Only one or less element in the first level.
+        print(len(shapes))
```
fchollet, line 384: Leftover from debugging?

PR #1146, opened 2020-05-25
enable and tested saving of categorical to numerical layer

haifeng-jin:

resolves #000

### DIFF
--- a/autokeras/keras_layers.py
+++ b/autokeras/keras_layers.py
@@ -76,4 +76,5 @@ class MultiColumnCategoricalEncoding(preprocessing.PreprocessingLayer):
 
 CUSTOM_OBJECTS = {
     'MultiColumnCategoricalEncoding': MultiColumnCategoricalEncoding,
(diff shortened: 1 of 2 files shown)


autokeras/keras_layers.py
```diff
@@ -76,4 +76,5 @@ def get_config(self):
 
 CUSTOM_OBJECTS = {
     'MultiColumnCategoricalEncoding': MultiColumnCategoricalEncoding,
+    'IndexLookup': index_lookup.IndexLookup,
```
fchollet, line 79: Consider using `StringLookup` and `IntegerLookup` respectively, instead of `IndexLookup`. The latter will not be exposed in the API

haifeng-jin: It seems they are only in 2.3. We are basing on 2.2 now. I will add an issue about this and about bump up to 2.3. We will add them when we upgrade to tf 2.3. Thanks.

PR #1148, opened 2020-05-25
Adding attention block

### LINKED ISSUE #723: Add recurrent attention to RNN Block
`feature_request`.

### Feature Description

The current implementation adds plain attention to the inputs or the final outputs. Need recurrent attention on each layer of the RNN model

### Reason

Recurrent attention captures chronologic patterns in the representations learned by each RNN layer.

achaar:
Feature Description

The current implementation adds plain attention to the inputs or the final outputs. Need recurrent attention on each layer of the RNN model

resolves #723

Previous code changes have been added to basic.py for attention in RNNBlock.


achaar:
The RNN attention block has been added. Although, I am not sure how we can prevent it from being used in Bidirectional RNNs.

autokeras/blocks/basic.py
```diff
@@ -154,6 +154,54 @@ def build(self, hp, inputs=None):
...
+    # Arguments
+        return_sequences: Boolean. Whether to return the last output in the
+            output sequence, or the full sequence. Defaults to False.
+    """
+
+    def __init__(self,
+                 return_sequences: bool = False,
+                 **kwargs):
+        super().__init__(**kwargs)
+        self.return_sequences = return_sequences
```
haifeng-jin, line 169: I don't think we need this. If it is False, there is no need to use this attention block.

autokeras/blocks/basic.py
```diff
@@ -154,6 +154,48 @@ def build(self, hp, inputs=None):
...
+
+    inputs: List of the following tensors:
+      * query: Query `Tensor` of shape `[batch_size, Tq, dim]`.
+      * value: Value `Tensor` of shape `[batch_size, Tv, dim]`.
+      * key: Optional key `Tensor` of shape `[batch_size, Tv, dim]`. If not
+        given, will use `value` for both `key` and `value`, which is the
+        most common case.
+
+    Output shape:
+    Attention outputs of shape `[batch_size, Tq, dim]`.
```
haifeng-jin, line 168: I just realized it would be better to put this into the build function's docstring, as # Arguments and # Returns

achaar:
I can present the attention block in tomorrow's meeting. But for it to work with the RNN block, the RNN block also needs to be modified. Currently, it does not return the hidden state of the RNN, which is required for the attention block.

haifeng-jin:
@achaar 
I think there are two kinds of attentions.
One is Attention layer.
The other on is AdditiveAttention layer.
The latter one is more commonly used.
The original implementation in another pull request is about the latter one.

So we may discuss the two attention mechanisms and decide which one to use. Or may be we can use both.
You can find the paper links of these two in https://buomsoo-kim.github.io/attention/2020/01/01/Attention-mechanism-1.md/#:~:text=Attention%20Mechanism%20in%20Neural%20Networks%20%2D%201.&text=Attention%20is%20arguably%20one%20of,a%20large%20amount%20of%20information

Thanks.

haifeng-jin:
@achaar I just consulted Keras team for how to do the attentions. It is not easy to implement. So let's just do the following.

1. Address the comments for moving the docstrings.
2. Use the hp to select between the two kinds of attention layers (Attention, AdditiveAttention).

Then this PR is good to be merged.
This Attention block will be used by users who are building their own search spaces with the blocks.

We will explore other ways to improve the performance of the text tasks.

Thanks.

autokeras/blocks/basic.py
```diff
@@ -154,6 +154,68 @@ def build(self, hp, inputs=None):
...
+        super().__init__(**kwargs)
+
+    def get_config(self):
+        config = super().get_config()
+        config.update({
+            'attention_type': self.attention_type})
+        return config
+
+    def build(self, hp, inputs=None):
+        '''
```
haifeng-jin, line 179: Should use """

autokeras/blocks/basic.py
```diff
@@ -154,6 +154,68 @@ def build(self, hp, inputs=None):
...
+
+        inputs: List of the following tensors:
```
haifeng-jin, line 181: Put these under # Arguments

autokeras/blocks/basic.py
```diff
@@ -154,6 +154,68 @@ def build(self, hp, inputs=None):
         return output_node
 
 
+class AttentionBlock(block_module.Block):
+    """The Attention Block. Luong's Multiplicative Style Attention.
+
+    # Arguments
+        attention_type: String. 'multiplicative' or 'additive'.
+        If left unspecified, it will be tuned automatically.
```
haifeng-jin, line 162: indent

autokeras/blocks/basic.py
```diff
@@ -154,6 +154,68 @@ def build(self, hp, inputs=None):
...
+        '''
+
+        inputs: List of the following tensors:
+          * query: Query `Tensor` of shape `[batch_size, Tq, dim]`.
+          * value: Value `Tensor` of shape `[batch_size, Tv, dim]`.
+          * key: Optional key `Tensor` of shape `[batch_size, Tv, dim]`. If not
+            given, will use `value` for both `key` and `value`, which is the
+            most common case.
+
+        Output shape:
```
haifeng-jin, line 188: # Returns

autokeras/blocks/basic.py
```diff
@@ -154,6 +154,69 @@ def build(self, hp, inputs=None):
         return output_node
 
 
+class AttentionBlock(block_module.Block):
+    """The Attention Block. Luong's Multiplicative Style Attention.
+
+    # Arguments
+        attention_type: String. 'multiplicative' or 'additive'.
+            If left unspecified, it will be tuned automatically.
+
```
haifeng-jin, line 163: Remove this empty line.

autokeras/blocks/basic.py
```diff
@@ -154,6 +154,69 @@ def build(self, hp, inputs=None):
...
+            inputs: List of the following tensors:
+              * query: Query `Tensor` of shape `[batch_size, Tq, dim]`.
+              * value: Value `Tensor` of shape `[batch_size, Tv, dim]`.
+              * key: Optional key `Tensor` of shape `[batch_size, Tv, dim]`. If not
+                given, will use `value` for both `key` and `value`, which is the
+                most common case.
+
+        # Returns:
+            Attention outputs of shape `[batch_size, Tq, dim]`.
+
```
haifeng-jin, line 191: Remove this empty line.

autokeras/blocks/basic.py
```diff
@@ -154,6 +154,69 @@ def build(self, hp, inputs=None):
...
+            'additive': layers.AdditiveAttention
+        }
+        if len(inputs) == 2 or len(inputs) == 3:
+            # Either [query, value] or [query, value, key] provided
+            for in_vec in inputs:
+                if len(in_vec.shape) != 3:
+                    raise ValueError(
+                        'Expect 3 dimensions for each of query, value, (or key) '
+                        'but got {shape}'.format(shape=in_vec.shape)
+                    )
```
haifeng-jin, line 209: I expect we have this check in the Keras Attention layers. You may have a try to see the exception the Attention and AdditiveAttention raises. If they have, we can remove this check.

autokeras/blocks/basic.py
```diff
@@ -154,6 +154,69 @@ def build(self, hp, inputs=None):
...
+            output_node = attention_layers[attention_type](inputs)
+        else:
+            raise ValueError(
+                'Expect the input to be [query, value] or [query, value, key]'
+                'for the Attention Layer, '
+                'but got {shape}'.format(shape=inputs.shape))
```
haifeng-jin, line 215: The same for this error. I expect the keras layer checked this.

autokeras/blocks/basic.py
```diff
@@ -154,6 +154,69 @@ def build(self, hp, inputs=None):
         return output_node
 
 
+class AttentionBlock(block_module.Block):
+    """The Attention Block. Luong's Multiplicative Style Attention.
```
haifeng-jin, line 158: This description needs to add the AddtiveAttention layer.
Just change this line to "Using Attention or AdditiveAttention layer in Keras."

autokeras/blocks/basic.py
haifeng-jin: If they don't have the exception. We need to check each tensor respectively and give the expected dimensions in details like [batch_size, Tq, dim] in the error message.

PR #1164, opened 2020-06-02
return one-hot instead of probability for multi-label

### LINKED ISSUE #1122: Multi-label classification `predict` method return probabilities rather than calibrated labels.
### Bug Description

Multi-label classification `predict()` method return probabilities rather than calibrated labels.

### Bug Reproduction
https://github.com/datamllab/automl-in-action-notebooks/blob/master/3.3.4-Task-API-Multi-label.ipynb

### Setup Details
 - OS type and version:
 - Python: 
 - autokeras: 1.0.2
 - keras-tuner:
 - scikit-learn:
 - numpy:
 - pandas:
 - tensorflow: 2.1.0

haifeng-jin:

resolves #1122

### DIFF
--- a/autokeras/adapters/output_adapter.py
+++ b/autokeras/adapters/output_adapter.py
@@ -78,6 +78,7 @@ class ClassificationHeadAdapter(HeadAdapter):
             dataset = dataset.values.reshape(-1, 1)
 
         # If encoded.
+        # TODO: support raw string labels for multi-label.
         if len(dataset.flatten()) != len(dataset):
             if self.num_classes:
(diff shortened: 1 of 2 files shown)


haifeng-jin:
> Why one-hot as opposed to probabilities or, says, lists of integers?

Our logic is output is in the same format as the user provided targets.
The targets for multi-label classification has to be one-hot encoded for now.

In the future, we will support lists of integers as training targets.
If the user provide lists of integers as training targets, we will return lists of integers as prediction results.

PR #1186, opened 2020-06-10
add tuner to task api args

haifeng-jin:

resolves #000

### DIFF
--- a/autokeras/tasks/image.py
+++ b/autokeras/tasks/image.py
@@ -2,6 +2,7 @@ from pathlib import Path
 from typing import List
 from typing import Optional
 from typing import Tuple
+from typing import Type
 from typing import Union
 
 import tensorflow as tf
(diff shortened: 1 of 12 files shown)


autokeras/tasks/image.py
```diff
@@ -41,6 +43,10 @@ class ImageClassifier(SupervisedImagePipeline):
             AutoModel in the current directory.
         objective: String. Name of model metric to minimize
             or maximize, e.g. 'val_accuracy'. Defaults to 'val_loss'.
+        tuner: String or subclass of AutoTuner. If use string, it should be one of
```
fchollet, line 46: "If string,"

autokeras/tasks/image.py
```diff
@@ -41,6 +43,10 @@ class ImageClassifier(SupervisedImagePipeline):
...
+            'greedy', 'bayesian', 'hyperband' or 'random'. It can also be a subclass
+            of AutoTuner. If left unspecified, it will use a tuner specially designed
+            for the task.
```
fchollet, line 49: This last sentence could be more detailed: how is the decision made?

PR #1190, opened 2020-06-12
adding attention block

haifeng-jin:

resolves #000

### DIFF
--- a/autokeras/blocks/basic.py
+++ b/autokeras/blocks/basic.py
@@ -154,6 +154,56 @@ class RNNBlock(block_module.Block):
         return output_node
 
 
+class Attention(block_module.Block):
+    """Apply Attention or AdditiveAttention layer in Keras.
+
+    Please refer to the documentationof the two layers for more details.
+    https://keras.io/api/layers/attention_layers/
+
+    # Arguments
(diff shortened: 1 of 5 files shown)


autokeras/blocks/basic.py
```diff
@@ -154,6 +154,56 @@ def build(self, hp, inputs=None):
         return output_node
 
 
+class Attention(block_module.Block):
```
fchollet, line 157: In what end-to-end workflow is this intended to be used?

autokeras/blocks/basic.py
```diff
@@ -154,6 +154,56 @@ def build(self, hp, inputs=None):
...
+    """Apply Attention or AdditiveAttention layer in Keras.
+
+    Please refer to the documentationof the two layers for more details.
```
fchollet, line 160: It would be useful to have a code example to illustrate usage in the docstring.

autokeras/blocks/basic.py
haifeng-jin: We don't have an end-to-end workflow for it now. We implemented it by accident. I think it is potentially use by other blocks to enlarge the search space.

haifeng-jin:
This PR is temporarily closed until we find a good use case for these layers.

PR #1211, opened 2020-06-26
setup.py: Move tensorflow to extras_require

utkarshgupta137:
Move tensorflow requirement to extras_require so that autokeras can be installed if tensorflow-gpu is installed (as in tensorflow/tensorflow:latest-gpu).
Following the practice mentioned in https://github.com/tensorflow/tensorflow/issues/7166

TODO: Update other docs & Dockerfile
Created pull request to discuss the approach.

Docs will have to be updated to reflect autokeras[tf] & autokeras[tf_gpu]
I think 2 different docker images should be available - one for gpu & one without

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,7 +20,6 @@ setup(
     keywords=['AutoML', 'Keras'],
     install_requires=[
         'packaging',
-        'tensorflow>=2.2.0',
         'scikit-learn',
         'numpy',
         'pandas',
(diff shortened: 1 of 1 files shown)


utkarshgupta137:
I'm using autokeras on top of tensorflow docker images, which currently has tensorflow-gpu installed. When pip install autokeras is done, it tries to install tensorflow instead of detecting tensorflow-gpu.

PR #1218, opened 2020-06-27
bump tf version to 2.3.0

haifeng-jin:

resolves #000

### DIFF
--- a/autokeras/keras_layers.py
+++ b/autokeras/keras_layers.py
@@ -1,18 +1,13 @@
-import inspect
-
 import tensorflow as tf
 from tensorflow.keras.layers.experimental import preprocessing
-from tensorflow.python.keras.layers.preprocessing import index_lookup
 from tensorflow.python.util import nest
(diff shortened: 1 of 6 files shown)


autokeras/keras_layers.py
```diff
@@ -1,18 +1,13 @@
...
-CombinerPreprocessingLayer = inspect.getmro(preprocessing.Normalization)[1]
-Combiner = inspect.getmro(preprocessing.Normalization()._combiner.__class__)[1]
-
 INT = 'int'
 NONE = 'none'
 ONE_HOT = 'one-hot'
 
 
+@tf.keras.utils.register_keras_serializable()
 class MultiColumnCategoricalEncoding(preprocessing.PreprocessingLayer):
```
fchollet, line 11: Note that the layer is named "CategoryEncoding", so it would be more consistent to reuse the same name here, e.g. "MultiCategoryEncoding"

PR #1231, opened 2020-07-13
add pretrained weights and cifar10 example

haifeng-jin:

resolves #000

1. Added pretrained weights option for Xception and ResNet.

2.  Using keras applications for corresponding blocks,  and removed unsupported args.
This change will minimize the search space to improve efficiency.

3. Added CIFAR10 example. Can achieve good performance with the pretrained model.
In tuning the performance, minor changes are made to ImageAugmentation and learning rate search space.


autokeras/graph.py
```diff
@@ -262,9 +262,19 @@ def _get_loss(self):
...
-                              default='adam')
+        optimizer_name = hp.Choice('optimizer',
+                                   ['adam', 'adadelta', 'sgd'],
+                                   default='adam')
+        learning_rate = hp.Choice('learning_rate',
+                                  [1e-1, 1e-2, 1e-3, 1e-4, 1e-5],
+                                  default=1e-3)
+
+        if optimizer_name == 'adam':
+            optimizer = tf.keras.optimizers.Adam(lr=learning_rate)
```
fchollet, line 273: `lr` is a legacy alias, use `learning_rate`

PR #1242, opened 2020-07-19
Pre-trained Bert

achaar:

resolves #000

Adding a pre-trained Bert model to be used as an embedding encoder.


autokeras/graph.py
```diff
@@ -306,6 +309,21 @@ def _compile_keras_model(self, hp, model):
             optimizer = tf.keras.optimizers.Adadelta(learning_rate=learning_rate)
         elif optimizer_name == "sgd":
             optimizer = tf.keras.optimizers.SGD(learning_rate=learning_rate)
+        elif optimizer_name == "adam_weight_decay":
+            # Set up epochs and steps
+            epochs = 3
```
achaar, line 314: How to set these values?

autokeras/graph.py
```diff
@@ -306,6 +309,21 @@ def _compile_keras_model(self, hp, model):
...
+            batch_size = 16
+            # eval_batch_size = 32
```
achaar, line 316: How to set the values of epochs and batch_size?

haifeng-jin: Currently, batch size cannot be a hyper parameter and cannot be changed according to the model used. Is it important for BERT model to tune it?

achaar: As below, it is used to calculate the warmup_steps and num_train_steps. If we let it be, those values would be wrong but it would still work. Also, I would need the train_data_size. Currently, I have just used the default values. It works.

PR #1310, opened 2020-08-24
add normalize to structured data block

haifeng-jin:

resolves #000

### DIFF
--- a/autokeras/blocks/wrapper.py
+++ b/autokeras/blocks/wrapper.py
@@ -181,14 +181,21 @@ class StructuredDataBlock(block_module.Block):
     # Arguments
         categorical_encoding: Boolean. Whether to use the CategoricalToNumerical to
             encode the categorical features to numerical features. Defaults to True.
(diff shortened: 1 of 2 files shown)


autokeras/blocks/wrapper.py
```diff
@@ -181,14 +181,21 @@ class StructuredDataBlock(block_module.Block):
     # Arguments
         categorical_encoding: Boolean. Whether to use the CategoricalToNumerical to
             encode the categorical features to numerical features. Defaults to True.
+        normalize: Boolean. Whether to normalize the features.
```
fchollet, line 184: This is feature-wise normalization. Do you think we will ever need to introduce sample-wise normalization as well?

haifeng-jin: I think BatchNorm can have similar effects for our current tasks. To my understanding, sample-wise or instance normalization are mainly used in generative models. We can add it when we do generative models.

PR #1332, opened 2020-09-06
Feature basic block

### LINKED ISSUE #1318: filter size as a parameter for convblock
hi, 

May I know why the filter size is not available as a parameter for the convblock? Could we have this feature instead of hardcoded values? Even for the DenseBlock too.

CuriousSingularity:
A custom kernel size, filter size and the unit size can be passed as arguments.

resolves #1318

Feature addition - A custom kernel size, filter size and the unit size can be passed as arguments.

### DIFF
--- a/autokeras/blocks/basic.py
+++ b/autokeras/blocks/basic.py
@@ -196,7 +196,7 @@ class ConvBlock(block_module.Block):
     """Block for vanilla ConvNets.
 
     # Arguments
-        kernel_size: Int. If left unspecified, it will be tuned automatically.
+        kernel_size: List. If left unspecified, it will be tuned automatically.
         num_blocks: Int. The number of conv blocks, each of which may contain
             convolutional, max pooling, dropout, and activation. If left unspecified,
             it will be tuned automatically.
@@ -213,7 +213,7 @@ class ConvBlock(block_module.Block):
 
     def __init__(
         self,
-        kernel_size: Optional[int] = None,
+        kernel_size: Optional[list] = None,
         num_blocks: Optional[int] = None,
         num_layers: Optional[int] = None,
         max_pooling: Optional[bool] = None,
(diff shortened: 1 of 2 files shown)


autokeras/blocks/basic.py
```diff
@@ -45,6 +45,7 @@ class DenseBlock(block_module.Block):
     # Arguments
         num_layers: Int. The number of Dense layers in the block.
             If left unspecified, it will be tuned automatically.
+        num_units: List. If left unspecified, it will be tuned automatically.
```
haifeng-jin, line 48: We need more descriptions here.

autokeras/blocks/basic.py
```diff
@@ -54,12 +55,14 @@ class DenseBlock(block_module.Block):
     def __init__(
         self,
         num_layers: Optional[int] = None,
+        num_units: Optional[list] = None,
```
haifeng-jin, line 58: We won't allow a list for now. It will be a feature in the future for user to pass hyperparameter instances.

autokeras/blocks/basic.py
```diff
@@ -188,12 +200,13 @@ class ConvBlock(block_module.Block):
...
 
     # Arguments
-        kernel_size: Int. If left unspecified, it will be tuned automatically.
+        kernel_size: List. If left unspecified, it will be tuned automatically.
         num_blocks: Int. The number of conv blocks, each of which may contain
             convolutional, max pooling, dropout, and activation. If left unspecified,
             it will be tuned automatically.
         num_layers: Int. The number of convolutional layers in each block. If left
             unspecified, it will be tuned automatically.
+        filter_size: List. If left unspecified, it will be tuned automatically.
```
haifeng-jin, line 209: the same here.

autokeras/blocks/basic.py
```diff
@@ -188,12 +200,13 @@ class ConvBlock(block_module.Block):
     """Block for vanilla ConvNets.
 
     # Arguments
-        kernel_size: Int. If left unspecified, it will be tuned automatically.
+        kernel_size: List. If left unspecified, it will be tuned automatically.
```
haifeng-jin, line 203: the same here. so kernel size should be left as it is.

haifeng-jin (changes requested):
Thanks for the PR! please see the comments.

haifeng-jin:
I am closing this pr to create another pr with your commits.
You will still be recognized as a contributor.

Thanks.

haifeng-jin:
The new PR is #1408

PR #1396, opened 2020-10-22
Add links to tf.keras.Model.fit for keras.Model.fit references in documentation

sandeepmistry:

N/A

This pull request proposes changes to improve the documentation, by adding a link to https://www.tensorflow.org/api_docs/python/tf/keras/Model#fit in the documentation for when `keras.Model.fit` is mentioned.

It could be expanded to add other links for Keras classes/methods, but wanted to get a general feel for if the maintainers would be open to changes like this.

Here is a preview of how the documentation would look:

Screen Shot 2020-10-22 at 5 09 18 PM

### DIFF
--- a/autokeras/auto_model.py
+++ b/autokeras/auto_model.py
@@ -241,7 +241,7 @@ class AutoModel(object):
                 validation data should be the same as the training data.
                 The best model found would be fit on the training dataset without the
                 validation data.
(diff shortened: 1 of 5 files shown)


haifeng-jin (approved):
Thanks for the PR! This is really helpful. I will update the website.

PR #1401, opened 2020-10-25
Fix example in BertBlock comment to change import and parameter naming

pavanagrawal123:
The comment example for BertBlock previously referred to paths that were no longer in the package. Further, the name of the parameter was not lining up with what is used in the package now. I have corrected the comment so people following the comments will not run into any issues.

### DIFF
--- a/autokeras/blocks/basic.py
+++ b/autokeras/blocks/basic.py
@@ -789,11 +789,11 @@ class BertBlock(block_module.Block):
     ```python
         # Using the Transformer Block with AutoModel.
         import autokeras as ak
-        from autokeras import BERTBlock
+        from autokeras.blocks import BertBlock
         from tensorflow.keras import losses
 
         input_node = ak.TextInput()
-        output_node = BERTBlock(max_seq_len=128)(input_node)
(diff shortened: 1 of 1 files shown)


autokeras/blocks/basic.py
```diff
@@ -789,11 +789,11 @@ class BertBlock(block_module.Block):
     ```python
         # Using the Transformer Block with AutoModel.
         import autokeras as ak
-        from autokeras import BERTBlock
+        from autokeras.blocks import BertBlock
```
haifeng-jin, line 792: Please remove the .blocks since it can be directly imported from autokeras with `from autokeras import BertBlock`.
https://github.com/keras-team/autokeras/blob/master/autokeras/__init__.py#L16

haifeng-jin (changes requested):
Thanks for the PR!

autokeras/blocks/basic.py
pavanagrawal123: Done! Sorry about that -- I should have caught that.

haifeng-jin (approved):
LGTM

PR #1411, opened 2020-11-02
Adding typed info for all classes

### LINKED ISSUE #918: Adding types information on the API surface.

dineshkumarsarangapani:

resolves #918 

Added typed info for almost all APIs.

### DIFF
--- a/autokeras/tuners/greedy.py
+++ b/autokeras/tuners/greedy.py
@@ -209,7 +209,7 @@ class Greedy(tuner_module.AutoTuner):
         max_trials: int = 10,
         initial_hps: Optional[List[Dict[str, Any]]] = None,
         seed: Optional[int] = None,
-        hyperparameters: kerastuner.HyperParameters = None,
+        hyperparameters: Optional[kerastuner.HyperParameters] = None,
         tune_new_entries: bool = True,
         allow_new_entries: bool = True,
         **kwargs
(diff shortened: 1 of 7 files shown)


autokeras/tuners/greedy.py
```diff
@@ -200,14 +204,14 @@ def _generate_hp_values(self, hp_names):
...
-        max_trials=10,
-        initial_hps=None,
-        seed=None,
-        hyperparameters=None,
-        tune_new_entries=True,
-        allow_new_entries=True,
+        hypermodel: kerastuner.HyperModel,
+        objective: str = "val_loss",
+        max_trials: int = 10,
+        initial_hps: Optional[List[Dict[str, Any]]] = None,
```
dineshkumarsarangapani, line 210: @haifeng-jin I was not sure on the type for initial_hps the list of dict has str as Key but was able to take Any (bool, float, int, str) as value. Hope you are find with this.

dineshkumarsarangapani:
@haifeng-jin This PR is ready for review. Please review.

autokeras/tuners/greedy.py
haifeng-jin: It think it looks good for now at least. If we have other use cases, we can change it later. Thanks.

haifeng-jin (approved):
Thanks for the PR! LGTM

PR #1418, opened 2020-11-05
Switch Docker example in to haifengjin/autokeras

chaselee:
This pull request updates the documentation for Docker. The previously referenced image is out of date. This image is only 18 days old and should be a more official source.

### DIFF
--- a/docs/templates/docker.md
+++ b/docs/templates/docker.md
@@ -6,7 +6,7 @@
 The following command download Auto-Keras docker image to your machine.  
 
 ```
-docker pull garawalid/autokeras:latest
+docker pull haifengjin/autokeras:latest
 ```
 
 Image releases are tagged using the following format:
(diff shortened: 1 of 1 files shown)


chaselee:
May need to tweak some commands - only swapped the image reference thus far.

cakemountain:
Also worth noting that documentation on the `docker run` command needs to be updated as well. In the new version the volumes are mounted differently. Whereas this worked with the previous image:

```
docker run -it -v "$(pwd)":/app --shm-size 2G garawalid/autokeras:latest python main.py
```

The new docker image...

```
docker run -it -v "$(pwd)":/app --shm-size 2G haifengjin/autokeras:latest python main.py
```

Fails:

```
python: can't open file './main.py': [Errno 2] No such file or directory
```

haifeng-jin (approved):
Thanks for the PR! LGTM

haifeng-jin:
@cakemountain Thank you! I have created an issue. #1422

PR #1449, opened 2020-11-23
Updated keras tuner version, removed 1.0.2rc4 suffix because the lates…

lusob:
Updated keras tuner installation url to the latest release, removed the @1.0.2rc4 suffix from the url because the final release is already available.


haifeng-jin (approved):
Thank you for the PR! LGTM

PR #1465, opened 2020-12-16
Removed Sequence imports not used

lusob:

Just remove some unused imports

Removed Sequence module imports from examples and docs because it wasn't being used


haifeng-jin (approved):
Thanks for the PR!

PR #1472, opened 2020-12-21
Simple path fix in docs .md file

alessando-guida:
fix the path in the example pointing to the mnist.py file. 

problem: The docker container can't find mnist.py as it is mounted under the ```/app``` folder


haifeng-jin (approved):
Thanks for the PR!

PR #1506, opened 2021-02-16
Ideas for Refactoring

Anselmoo:
Dear AutoKeras teams, this PR should provide ideas for code refactoring, which are bump in one single PR


autokeras/blocks/preprocessing.py
```diff
@@ -211,9 +211,9 @@ def build(self, hp, inputs=None):
             flip_mode = ""
         elif horizontal_flip and vertical_flip:
             flip_mode = "horizontal_and_vertical"
-        elif horizontal_flip and not vertical_flip:
+        elif horizontal_flip:
             flip_mode = "horizontal"
-        elif not horizontal_flip and vertical_flip:
+        else:
```
haifeng-jin, line 216: The redundancy here is for readability.
I prefer to leave it as it was.

haifeng-jin (changes requested):
Thanks for the PR! After addressing the comments, the PR should be good to merge.

autokeras/blocks/preprocessing.py
Anselmoo: @haifeng-jin I understand that, then better close the PR

Anselmoo: @haifeng-jin sorry misunderstanding, you mean the preprocessing should stay as it is and then you are fine?

haifeng-jin (approved):
LGTM

PR #1508, opened 2021-02-17
Fix docs of Customized Model

yurimo36:
fix `tf.python.util.nest.flatten(inputs)[0]` -> `tf.nest.flatten(inputs)[0]`

reference:
https://www.tensorflow.org/api_docs/python/tf/nest/flatten


yurimo36:
@googlebot I signed it!

haifeng-jin (approved):
Thanks for the PR!

PR #1514, opened 2021-02-20
Transformers Block Choice

mandalbiswadip:
Fixes a part of #1420 

Enabling hyperparameter choices for the Transformer Block


haifeng-jin (approved):
Thanks for the PR!

PR #1520, opened 2021-03-07
Time Series Forecaster basic Tutorial

abgese:

resolves #000

### DIFF
--- a/docs/py/timeseries_forecaster.py
+++ b/docs/py/timeseries_forecaster.py
@@ -18,38 +18,76 @@ The first step is to prepare your data. Here we use the [UCI Airquality dataset]
 """
 
 dataset = tf.keras.utils.get_file(
-    fname='AirQualityUCI.csv',
+    fname="AirQualityUCI.csv",
     origin="https://archive.ics.uci.edu/ml/machine-learning-databases/00360/"
     "AirQualityUCI.zip",
-    extract=True
+    extract=True,
 )
 
-dataset = pd.read_csv(dataset, sep=';')
+dataset = pd.read_csv(dataset, sep=";")
 dataset = dataset[dataset.columns[:-2]]
 dataset = dataset.dropna()
-dataset = dataset.replace(',', '.', regex=True)
+dataset = dataset.replace(",", ".", regex=True)
 
 val_split = int(len(dataset) * 0.7)
 data_train = dataset[:val_split]
 validation_data = dataset[val_split:]
 
-data_x = data_train[['CO(GT)', 'PT08.S1(CO)', 'NMHC(GT)', 'C6H6(GT)',
-                     'PT08.S2(NMHC)', 'NOx(GT)', 'PT08.S3(NOx)', 'NO2(GT)',
-                     'PT08.S4(NO2)', 'PT08.S5(O3)', 'T', 'RH']].astype('float64')
+data_x = data_train[
+    [
+        "CO(GT)",
+        "PT08.S1(CO)",
+        "NMHC(GT)",
+        "C6H6(GT)",
+        "PT08.S2(NMHC)",
+        "NOx(GT)",
+        "PT08.S3(NOx)",
+        "NO2(GT)",
+        "PT08.S4(NO2)",
+        "PT08.S5(O3)",
+        "T",
+        "RH",
+    ]
+].astype("float64")
 
-data_x_val = validation_data[['CO(GT)', 'PT08.S1(CO)', 'NMHC(GT)', 'C6H6(GT)',
-                              'PT08.S2(NMHC)', 'NOx(GT)', 'PT08.S3(NOx)', 'NO2(GT)',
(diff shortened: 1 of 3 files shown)


docs/py/timeseries_forecaster.py
```diff
@@ -0,0 +1,67 @@
...
+
+data_x_val = validation_data[['CO(GT)', 'PT08.S1(CO)', 'NMHC(GT)', 'C6H6(GT)',
+                              'PT08.S2(NMHC)', 'NOx(GT)', 'PT08.S3(NOx)', 'NO2(GT)',
+                              'PT08.S4(NO2)', 'PT08.S5(O3)', 'T', 'RH']].astype(
+                                  'float64')
+
+data_y = data_train['AH'].astype('float64')
+
+data_y_val = validation_data['AH'].astype('float64')
+
```
haifeng-jin, line 45: It is good to print the shape of the data here.

docs/py/timeseries_forecaster.py
```diff
@@ -0,0 +1,67 @@
...
+    predict_from=predict_from,
+    predict_until=predict_until,
+    max_trials=1,
+    objective='val_loss'
+)
+
+clf.fit(x=data_x, y=data_y, validation_data=(data_x_val, data_y_val), batch_size=30,
+        epochs=10)
+
+predictions = clf.predict(data_x_val)
```
haifeng-jin, line 67: print the shape of the `predictions`.
Also call evaluation to see and print the result with a readable message `'...'.format()`

haifeng-jin (changes requested):
Thanks for the PR! Please sign the Google CLA following the instructions commented by the bot on this PR. Also, try to run the formatted to fix the CI.

docs/py/timeseries_forecaster.py
iamianM: my predictions array looks like this: `array([], shape=(0, 1), dtype=float32)`
Have you encountered this issue before?

abgese: Figured out the issue here. In the predict function, the y_pred is sliced to only output the predictions from "train_len + predict_from" to "train_len + predict_until". Therefore, if we use data_x_val(which has length less than the train_len) the function will return an empty array. I'll try to figure out what makes the most sense here.

Note: Another thing that might cause some issues with the slicing is that the length of the y_pred before slicing is equal to "test_len - look_back + 1". Sp a very large lookback might result in an empty prediction array again.

abgese:
@googlebot I signed it!

docs/py/timeseries_forecaster.py
abgese: So based on the how the api is expected to be used we are currently going to enforce, that the predict data needs to contain the original train data time steps along with the time steps for which the prediction needs to be done.

stancsz (approved):
I ran your branch & it seem functional, thanks for the tutorial

docs/py/timeseries_forecaster.py
```diff
@@ -0,0 +1,78 @@
...
+    predict_from=predict_from,
+    predict_until=predict_until,
+    max_trials=1,
+    objective='val_loss'
+)
+# Train the TimeSeriesForecaster with train data
+clf.fit(x=data_x, y=data_y, validation_data=(data_x_val, data_y_val), batch_size=32,
+        epochs=10)
+# Predict with the best model(includes original training data).
+predictions = clf.predict(data_x_test)
```
stancsz, line 75: the returned predictions are min-max scaled, not of the original scale. Is this intended?

stancsz (changes requested):

docs/py/timeseries_forecaster.py
abgese: I'm sorry could you clarify what you mean?

stancsz: image

image
@abgese  apologies, I have a use case that resulted in numbers like so. I thought they are min-max scaled, but really they're not. I think I must be using the code incorrectly. 
👀 & trying to find out what I'm doing wrong. 

please ignore this and let me move this conversation elsewhere.

docs/py/timeseries_forecaster.py
```diff
@@ -0,0 +1,78 @@
...
+
+print(data_x.shape)  # (6549, 12)
+print(data_y.shape)  # (6549,)
+
+"""
+The second step is to run the [TimeSeriesForecaster](/time_series_forecaster).
+As a quick demo, we set epochs to 10.
+You can also leave the epochs unspecified for an adaptive number of epochs.
+"""
+import autokeras as ak
```
stancsz, line 59:
```
E402 module level import not at top of file
```
the import of `autokeras` should be place on top of the file
context:
https://www.flake8rules.com/rules/E402.html

stancsz (changes requested):
```
E402 module level import not at top of file
```
the import of `autokeras` should be place on top of the file
context:
https://www.flake8rules.com/rules/E402.html

stancsz (approved):

Issue #1423, opened 2020-11-07
Return History


TKassis:
### Feature Description
In Keras it's possible to plot validation loss by accessing it through model.history, but Auto-Keras does not return an object containing this when I call .fit. What's the best way to access historical metrics for the best model?

gousemd73:
Yes, It would be very nice if can able plot the losses and accuracies that we recorded during the training.

haifeng-jin:
May be you can try `auto_model.tuner.results_summary()`? We will have a better solution in the next few months by improving Keras Tuner.

hanzigs:
Yes, I am waiting for this as well, Thanks

pa-blan:
@haifeng-jin You can already use Kerastuner to create the history of the final model, why not in autokeras?

SuiAn-97:
Waiting for this.

lwq-star:
What's the way to access historical metrics? I still receive this error.
AttributeError: 'StructuredDataRegressor' object has no attribute 'history'
`reg = ak.StructuredDataRegressor(overwrite=True,max_trials=3,
                                     metrics=[ta.metrics.RSquare(),
                                              tf.keras.metrics.RootMeanSquaredError(),"mse"])
 reg.fit(x, y, epochs=10)
 history = reg.history`

lwq-star:
What's the way to access historical metrics? I still receive this error.
AttributeError: 'StructuredDataRegressor' object has no attribute 'history'

```
reg = ak.StructuredDataRegressor(overwrite=True,max_trials=3, metrics=[ta.metrics.RSquare(), 
                                                                                                                      tf.keras.metrics.RootMeanSquaredError(),"mse"]) 
reg.fit(x, y, epochs=10) 
history = reg.history
```

BenUCL:
> 

This returns AttributeError: 'Functional' object has no attribute 'tuner'

PR #1561, opened 2021-05-13
Added Functionality for returning Model History

djokester:
Added functionality for returning a history object corresponding to the best model when the fit() function is called. #1423


autokeras/auto_model.py
```diff
@@ -252,6 +252,12 @@ def fit(
                 [keras.Model.fit](https://www.tensorflow.org/api_docs/python/tf/keras/Model#fit)
             **kwargs: Any arguments supported by
                 [keras.Model.fit](https://www.tensorflow.org/api_docs/python/tf/keras/Model#fit).
+
+        #Returns
```
haifeng-jin, line 256: This should be "# Returns"

haifeng-jin (approved):

PR #1658, opened 2021-12-23
fix docs of TimeSeriesForecaster

yurimarimo:

resolves #000

I fixed URLs with titles not being reflected because of code formatter.


haifeng-jin (approved):
The original link doesn't work because the "\n" would be changed to a space, which will break between [] and ().
Now it works well.

PR #1714, opened 2022-04-26
Update ImageAugmentation translation_factor hyperparameter to have a range

Neproxx:
Partial resolve for #1420

Added functionality for the user to specify a `hyperparameters.Choice` object for the `translation_factor` of `ImageAugmentation`. 

Note: although `translation_factor` accepts either a float or a tuple of floats, hyperparameters.Choice does not support tuples and is thus limited to a range of single floats.


haifeng-jin (approved):
Thanks for the PR!

PR #1724, opened 2022-05-14
Fixed: #1722 Run out of memory

### LINKED ISSUE #1722: `resnet_v2`-models can cause out of memory error
### Bug Description

In the case of using `resnet_v2`-models, an  **out-of-memory error** can be generated.

### Expected Behavior

Instead of crashing and generating an **out-of-memory error**, the currently used model should be skipped.

### Setup Details
 - OS type and version: Ubuntu 20
 - Python: 3.8
 - autokeras: latest version
 
### Additional context

See also discussion: #1721

Anselmoo:
Update utils.py and test by using `break` instead of `raise`

resolves #1722

According to discussion #1721, the`raise` will be replaced by `break` for running _out of memory_ in the case of `batch_size==1`
.

### DIFF
--- a/autokeras/utils/utils.py
+++ b/autokeras/utils/utils.py
@@ -94,15 +94,20 @@ def fit_with_adaptive_batch_size(model, batch_size, **fit_kwargs):
 def run_with_adaptive_batch_size(batch_size, func, **fit_kwargs):
     x = fit_kwargs.pop("x")
     validation_data = None
+    history = None
     if "validation_data" in fit_kwargs:
         validation_data = fit_kwargs.pop("validation_data")
     while batch_size > 0:
         try:
             history = func(x=x, validation_data=validation_data, **fit_kwargs)
             break
-        except tf.errors.ResourceExhaustedError as e:
+        except tf.errors.ResourceExhaustedError:
             if batch_size == 1:
-                raise e
+                print(
+                    "Not enough memory, reduced batch size is already set to 1. "
+                    "Current model will be skipped."
+                )
+                break
             batch_size //= 2
             print(
                 "Not enough memory, reduce batch size to {batch_size}.".format(
--- a/autokeras/utils/utils_test.py
+++ b/autokeras/utils/utils_test.py
@@ -53,19 +53,20 @@ def test_check_kt_version_error():
     )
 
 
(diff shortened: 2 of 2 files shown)


haifeng-jin (commented):
Thanks for the fix!

If history is None, it should trigger [this error](https://github.com/keras-team/keras-tuner/blob/master/keras_tuner/engine/tuner_utils.py#L283). To your test, does it trigger it?
We can first add a `if` to check its None, return an empty dict.

Check for empty dict in `Oracle.update_trial` to return [invalid status](https://github.com/keras-team/keras-tuner/blob/84eb5c40660fc0fa3745bcb04ee4b4693f80dc07/keras_tuner/engine/trial.py#L32).

I can make the change if you prefer?

Anselmoo:
> Thanks for the fix!
> 
> If history is None, it should trigger [this error](https://github.com/keras-team/keras-tuner/blob/master/keras_tuner/engine/tuner_utils.py#L283). To your test, does it trigger it? We can first add a `if` to check its None, return an empty dict.
> 
> Check for empty dict in `Oracle.update_trial` to return [invalid status](https://github.com/keras-team/keras-tuner/blob/84eb5c40660fc0fa3745bcb04ee4b4693f80dc07/keras_tuner/engine/trial.py#L32).
> 
> I can make the change if you prefer?

Will be the changes small, then I would prefer if you would do it because I have to first get into the code. Otherwise, I will take later a look.

haifeng-jin:
@Anselmoo I will make the fix then, it may take some time. Thanks.

haifeng-jin:
This is blocked by:
https://github.com/keras-team/keras-tuner/issues/704

autokeras/utils/utils.py
```diff
@@ -94,15 +94,20 @@ def fit_with_adaptive_batch_size(model, batch_size, **fit_kwargs):
...
         validation_data = fit_kwargs.pop("validation_data")
     while batch_size > 0:
         try:
             history = func(x=x, validation_data=validation_data, **fit_kwargs)
             break
-        except tf.errors.ResourceExhaustedError as e:
+        except tf.errors.ResourceExhaustedError:
             if batch_size == 1:
-                raise e
+                print(
```
auvipy, line 106: I am not sure if using print a right approach here?

Issue #1480, opened 2021-01-06
Fix Hyper-Parameters For AutoKeras Standard Method Search Spaces


WilliamJudge94:
### Feature Description
A way to fix certain aspects of the search space such as learning rate, optimizer, etc for every model created by the ImageClassifier.

### Reason
Allows a User to minimize their search space for the standard AutoKeras search methods.

haifeng-jin:
This feature will be enabled by #1494.

PR #1736, opened 2022-06-11
Fixed: #1494 Check optimizer and learning rate settings

### LINKED ISSUE #1494: add compile function to automodel
### Code Example

```python
```

jwliou:

Partial resolves #1494

In fact, this fix is more likely a resolving for #1480. (with implicit insertion of rmsprop and adadelta optimizers with default settings.)

This pull request is for checking if there already exist hyperparameter settings for optimizer and learning rate in current hp.
Original code will just overwrite any possible optimizer and learning rate settings already in hp so that every model needs to search from a preset optimizer and learning rate space even if you set these in elsewhere such as tuners/task_specific.py.

It could be better to serialize optimizer settings, but where to put the serialization operations will be the next problem.

### DIFF
--- a/autokeras/graph.py
+++ b/autokeras/graph.py
@@ -279,20 +279,33 @@ class Graph(keras_tuner.HyperModel, serializable.Serializable):
 
     def _compile_keras_model(self, hp, model):
         # Specify hyperparameters from compile(...)
-        optimizer_name = hp.Choice(
-            "optimizer",
-            ["adam", "sgd", "adam_weight_decay"],
-            default="adam",
-        )
+        # check if optimizer is set before applying default optimizer settings.
+        if "optimizer" in hp.values.keys():
+            optimizer_name = hp.get("optimizer")
+        else:
+            optimizer_name = hp.Choice(
+                "optimizer",
+                ["adam", "sgd", "adam_weight_decay"],
+                default="adam",
+            )
+            
         # TODO: add adadelta optimizer when it can optimize embedding layer on GPU.
-        learning_rate = hp.Choice(
-            "learning_rate", [1e-1, 1e-2, 1e-3, 1e-4, 2e-5, 1e-5], default=1e-3
-        )
+        # check if learning rate is set before applying default learning rate settings.
+        if "learning_rate" in hp.values.keys():
+            learning_rate = hp.get("learning_rate")
+        else:
+            learning_rate = hp.Choice(
+                "learning_rate", [1e-1, 1e-2, 1e-3, 1e-4, 2e-5, 1e-5], default=1e-3
+            )
 
         if optimizer_name == "adam":
             optimizer = keras.optimizers.Adam(learning_rate=learning_rate)
         elif optimizer_name == "sgd":
             optimizer = keras.optimizers.SGD(learning_rate=learning_rate)
+        elif optimizer_name == "rmsprop":
+            optimizer = keras.optimizers.RMSprop(learning_rate=learning_rate)
+        elif optimizer_name == "adadelta":
+            optimizer = keras.optimizers.Adadelta(learning_rate=learning_rate)
         elif optimizer_name == "adam_weight_decay":
             steps_per_epoch = int(self.num_samples / self.batch_size)
             num_train_steps = steps_per_epoch * self.epochs
--- a/autokeras/graph.py
+++ b/autokeras/graph.py
@@ -295,7 +295,8 @@ class Graph(keras_tuner.HyperModel, serializable.Serializable):
             learning_rate = hp.get("learning_rate")
         else:
             learning_rate = hp.Choice(
-                "learning_rate", [1e-1, 1e-2, 1e-3, 1e-4, 2e-5, 1e-5], default=1e-3
+                "learning_rate", [1e-1, 1e-2, 1e-3, 1e-4, 2e-5, 1e-5], 
+                default=1e-3,
             )
 
         if optimizer_name == "adam":


Anselmoo:
See also: #1734

jwliou:
Optimizer will be a larger block of settings including learning rate, instead of considering optimizer and learning rate separately.

autokeras/graph.py
```diff
@@ -279,20 +279,33 @@ def _get_loss(self):
...
         # Specify hyperparameters from compile(...)
-        optimizer_name = hp.Choice(
-            "optimizer",
-            ["adam", "sgd", "adam_weight_decay"],
-            default="adam",
-        )
+        # check if optimizer is set before applying default optimizer settings.
+        if "optimizer" in hp.values.keys():
+            optimizer_name = hp.get("optimizer")
+        else:
```
haifeng-jin, line 285: We don't need the check here. If the optimizer is already defined in the hp. It will not override its search space, but directly use the old one even it is different from the new one.
You may give a try.

autokeras/graph.py
```diff
@@ -279,20 +279,33 @@ def _get_loss(self):
...
+            )
+            
         # TODO: add adadelta optimizer when it can optimize embedding layer on GPU.
-        learning_rate = hp.Choice(
-            "learning_rate", [1e-1, 1e-2, 1e-3, 1e-4, 2e-5, 1e-5], default=1e-3
-        )
+        # check if learning rate is set before applying default learning rate settings.
+        if "learning_rate" in hp.values.keys():
+            learning_rate = hp.get("learning_rate")
+        else:
```
haifeng-jin, line 296: Same here, we don't need the check.

haifeng-jin (commented):
Thanks for the PR!
Please address the comments.
After that this PR should only contain the 2 added optimizers.
Please use `sh shell/format.sh` to format the code. Thanks.

autokeras/graph.py
jwliou: But if the hp comes with a certain default values for optimizer_name and learning_rate, 
such values will be overwritten in the so-called "search space."

In autokeras/tuners/task_specific.py, there exists such setting:

```
IMAGE_CLASSIFIER = [
    {
        "image_block_1/block_type": "vanilla",
        "image_block_1/normalize": True,
        # some irrelevant settings omitted for shorter reply.
        "classification_head_1/dropout": 0.5,
        "optimizer": "adam",
        "learning_rate": 1e-3,
    },
    {
        "image_block_1/block_type": "resnet",
        "image_block_1/normalize": True,
        # some irrelevant settings omitted for shorter reply.
        "classification_head_1/dropout": 0,
        "optimizer": "adam",
        "learning_rate": 1e-3,
    },
    {
        "image_block_1/block_type": "efficient",
        "image_block_1/normalize": True,
        # some irrelevant settings omitted for shorter reply.
        "classification_head_1/dropout": 0,
        "optimizer": "adam",
        "learning_rate": 2e-5,
    },
]
```

what will happen if I had modified learning rate to something like 2e-4, which is not within 
the default learning rate range [1e-1, 1e-2, 1e-3, 1e-4, 2e-5, 1e-5]?

modified code with checking will lead to...
learning_rate = hp.get("learning_rate") => get the learning_rate setting 2e-4.

but...
learning_rate = hp.Choice(
       "learning_rate", [1e-1, 1e-2, 1e-3, 1e-4, 2e-5, 1e-5], 
       default=1e-3,
)
can never get learning_rate setting to 2e-4, that is the problem.

The same story will go for optimizer name, but since the only allowed optimizer names 
are all listed before this pull request, previous statement might be worked in some sense, 
but such implicit problem still exists.

autokeras/graph.py
```diff
@@ -279,20 +279,34 @@ def _get_loss(self):
...
+            )
 
         if optimizer_name == "adam":
             optimizer = keras.optimizers.Adam(learning_rate=learning_rate)
         elif optimizer_name == "sgd":
             optimizer = keras.optimizers.SGD(learning_rate=learning_rate)
+        elif optimizer_name == "rmsprop":
+            optimizer = keras.optimizers.RMSprop(learning_rate=learning_rate)
+        elif optimizer_name == "adadelta":
+            optimizer = keras.optimizers.Adadelta(learning_rate=learning_rate)
```
Anselmoo, line 309: Is there any specific reason why only these two solvers are added?

https://keras.io/api/optimizers/

If adding `Adadelta`, why not adding `Adagrad`, too?

jwliou: In fact, this is not my major purpose to add optimizers of adam series. It may just because I had seen adadelta in the comments of codes.

For RMSprop, that is mainly for personal preference. Besides variations of adam optimizers, optimizers implemented in Keras are SGD, RMSprop, and Ftrl according to official document. I'm wondering if the last one may be preferred for some other programmer.

For more general request of custom optimizer, it still need to refer to my previous comment: 
Serialization of optimizer settings.
But such serialization will treat learning rate as one parameter within optimizer settings, 
and you will need to implement your own optimizer as subclass of Keras optimizer.
Further more, learning rate scheduler might also be considered.

A possible reference or example for serialization is loss or metric, which implement serialization in engine/head.py.
Optimizer may need to follow such step in long term, but it may not be suitable to implement in the same place as loss and metric since it is reasonable to assume one model/graph can only have one universal optimizer, but one model/graph may have multiple heads. Maybe I am wrong here.

and optimizer serialization may not only put extra codes in utils/type.py like this simple way:
```
# leave loss for example
from tensorflow.keras.losses import Loss

LossType = Union[str, Callable, Loss]

# just share the same type definition with loss
from tensorflow.keras.optimizers import Optimizer

OptimizerType = Union[str, Callable, Optimizer]
```

The true purpose of this request may need more effort to accomplish though.

autokeras/graph.py
haifeng-jin: So you intend to use this with `ImageClassifier()` API or similar? Would you please give a short code example to illustrate how the user would use this feature you contributed?

Thanks!

jwliou: Tested example as following:
```
import numpy as np
import matplotlib.pyplot as plt
import tensorflow as tf
from tensorflow.keras.datasets import cifar10
import autokeras as ak
from autokeras.tuners import greedy

# Custom setting of Image Classifier Tuner
MY_IMAGE_CLASSIFIER = [
    {
        "image_block_1/block_type": "resnet",
        "image_block_1/normalize": True,
        "image_block_1/augment": True,
        "image_block_1/image_augmentation_1/horizontal_flip": True,
        "image_block_1/image_augmentation_1/vertical_flip": True,
        "image_block_1/image_augmentation_1/contrast_factor": 0.0,
        "image_block_1/image_augmentation_1/rotation_factor": 0.0,
        "image_block_1/image_augmentation_1/translation_factor": 0.1,
        "image_block_1/image_augmentation_1/zoom_factor": 0.0,
        "image_block_1/res_net_block_1/pretrained": False,
        "image_block_1/res_net_block_1/version": "resnet50",
        "image_block_1/res_net_block_1/imagenet_size": True,
        "classification_head_1/spatial_reduction_1/reduction_type": "global_avg",
        "classification_head_1/dropout": 0,
        "optimizer": "adam",
        "learning_rate": 5e-4,
    },
    {
        "image_block_1/block_type": "efficient",
        "image_block_1/normalize": True,
        "image_block_1/augment": True,
        "image_block_1/image_augmentation_1/horizontal_flip": True,
        "image_block_1/image_augmentation_1/vertical_flip": False,
        "image_block_1/image_augmentation_1/contrast_factor": 0.0,
        "image_block_1/image_augmentation_1/rotation_factor": 0.0,
        "image_block_1/image_augmentation_1/translation_factor": 0.1,
        "image_block_1/image_augmentation_1/zoom_factor": 0.0,
        "image_block_1/efficient_net_block_1/pretrained": True,
        "image_block_1/efficient_net_block_1/version": "b3",
        "image_block_1/efficient_net_block_1/trainable": True,
        "image_block_1/efficient_net_block_1/imagenet_size": True,
        "classification_head_1/spatial_reduction_1/reduction_type": "global_avg",
        "classification_head_1/dropout": 0,
        "optimizer": "rmsprop",
        "learning_rate": 1e-4,
    },
    {
        "image_block_1/block_type": "efficient",
        "image_block_1/normalize": True,
        "image_block_1/augment": True,
        "image_block_1/image_augmentation_1/horizontal_flip": True,
        "image_block_1/image_augmentation_1/vertical_flip": False,
        "image_block_1/image_augmentation_1/contrast_factor": 0.0,
        "image_block_1/image_augmentation_1/rotation_factor": 0.0,
        "image_block_1/image_augmentation_1/translation_factor": 0.1,
        "image_block_1/image_augmentation_1/zoom_factor": 0.0,
        "image_block_1/efficient_net_block_1/pretrained": True,
        "image_block_1/efficient_net_block_1/version": "b4",
        "image_block_1/efficient_net_block_1/trainable": True,
        "image_block_1/efficient_net_block_1/imagenet_size": True,
        "classification_head_1/spatial_reduction_1/reduction_type": "global_avg",
        "classification_head_1/dropout": 0,
        "optimizer": "adam",
        "learning_rate": 5e-5,
    },
]

class MyImageClassifierTuner(greedy.Greedy):
    def __init__(self, **kwargs):
        super().__init__(initial_hps=MY_IMAGE_CLASSIFIER, **kwargs)

# Loading cifar dataset
(x_train, y_train), (x_test, y_test) = cifar10.load_data()

print(x_train.shape)
print(x_test.shape)

# Apply custom tuner in ImageClassifier to select from custom models
clf = ak.ImageClassifier(max_trials=3, tuner=MyImageClassifierTuner)

# Training, testing and print fitting results.
clf.fit(x_train, y_train, epochs=10, batch_size=16)
predicted = clf.predict(x_test[:10])

labels = ('airplane', 'automobile', 'bird', 'cat', 'deer', 'dog', 'frog', 'horse', 'ship', 'truck')

fig = plt.figure(figsize=(16, 6))
for i in range(10):
    ax = fig.add_subplot(2, 5, i + 1)
    ax.set_axis_off()
    plt.imshow(x_test[i])
    ax.set_title(f'Predicted: {labels[int(predicted[i])]},\nReal: {labels[int(y_test[i])]}')

    plt.tight_layout()
plt.show()

model = clf.export_model()
model.summary()
```

haifeng-jin: This is too complicated for other users. Let's just add the `optimizer` argument to `AutoModel` class.

jwliou: OK... I think such simplist example will show difference of this pull.
But if the original behavior still considered as better... I have nothing to say and I'll just create another branch for private use.

```py
import numpy as np
import matplotlib.pyplot as plt
import tensorflow as tf
from tensorflow.keras.datasets import cifar10
import autokeras as ak

# Loading cifar dataset
(x_train, y_train), (x_test, y_test) = cifar10.load_data()

print(x_train.shape)
print(x_test.shape)

# Change this 3 to 9, 12 or 27, the difference will be shown.
nc = 9
# clf = ak.ImageClassifier(max_trials=3)
clf = ak.ImageClassifier(max_trials=nc)

# Training, testing and print fitting results.
clf.fit(x_train, y_train, epochs=10, batch_size=16)
clf.tuner.results_summary(num_trials=nc)
```

autokeras/graph.py
djokester: Hey!
Would you be willing to work on adding functionality for passing loss and optimizer in the form of tunable hyper parameters as arguments for automodel?

jwliou: well, it is possible if overwriting optimizer can be avoided, but something weird happened for real code testing above.

PR #1768, opened 2022-08-31
1.0.21dev

haifeng-jin:

resolves #000


YASIRAQ:
hello . I want to do text classification by using auto-keras . but I want to specify word2vec as word embedding technique. how can  I do that ? thank you very much

PR #1782, opened 2022-09-28
fix: documentation allign code and comment

buti1021:


haifeng-jin (commented):
Thanks for the PR! Everything looks good. Would you please format the code with `shell/format.sh` so that the CI code style check would pass?

haifeng-jin (approved):

PR #1816, opened 2022-12-15
switch bert to KerasNLP implementation

haifeng-jin:
AutoKeras has switched to KerasNLP for BERT implementation without any problem.
Successfully removed 1500+ lines of code!
This PR is pending on KerasNLP 0.4.0 release before merging.

Due to the limitation of AutoKeras, we are using the `BertPreprocessor` as part of the main Keras model.
In the future, this should be in the `tf.data` pipeline instead.

cc: @mattdangerw, @jbischof, @chenmoneygithub


autokeras/blocks/basic.py
```diff
@@ -974,14 +976,11 @@ def from_config(cls, config):
     def build(self, hp, inputs=None):
         input_tensor = nest.flatten(inputs)[0]
 
-        tokenizer_layer = keras_layers.BertTokenizer(
-            max_sequence_length=utils.add_to_hp(self.max_sequence_length, hp)
-        )
-        output_node = tokenizer_layer(input_tensor)
-
-        bert_encoder = keras_layers.BertEncoder()
+        preset_name = "bert_base_en_uncased"
```
jbischof, line 979: Depending on your application and user base, does using an EN-only model make sense? We offer `"bert_base_multi_cased"` which is trained on 104 languages as well as much more powerful models such as `XLMRoberta` which have training data and tokenization optimized for multilingual settings.

jbischof (commented):
Very cool @haifeng-jin, excited for the collaboration!

autokeras/blocks/basic.py
haifeng-jin: For this PR, we try to keep it consistent as before. So we chose this preset.

The multi-language model is a great idea! In the future, we plan to let the user specify their language for `TextClassifier`, base on which AutoKeras will choose the right model and preset for them.

jbischof: Sure, happy to talk later! After BERT, most models were trained on a unified multilingual corpus. In fact, XLMRoberta is famous for showing that this is even better for EN-only. So with the right model you may not need to do this.

autokeras/blocks/basic.py
```diff
@@ -974,14 +976,14 @@ def from_config(cls, config):
...
+            sequence_length=utils.add_to_hp(self.max_sequence_length, hp),
         )
-        output_node = tokenizer_layer(input_tensor)
-
-        bert_encoder = keras_layers.BertEncoder()
+        bert_encoder = keras_nlp.models.BertBackbone.from_preset(preset_name)
 
-        output_node = bert_encoder(output_node)
-        bert_encoder.load_pretrained_weights()
+        output_node = tokenizer_layer(tf.reshape(input_tensor, [-1]))
```
mattdangerw, line 986: Why do you need a reshape here? What is the shape of the input tensor?

mattdangerw (approved):
This looks great to me! Excited to have autokeras use KerasNLP!

Had one question that was unclear to me.

autokeras/blocks/basic.py
haifeng-jin: The shape is `[batch_size, 1]`. The input tensor is not `["This is the first sentence.", "This is the second one."]`, but `[["This is the first sentence."], ["This is the second one."]]`. It has one more dimension.

jbischof:
KerasNLP 0.4 is released. Merge away!

jbischof (approved):
Excited for the collaboration!

mattdangerw:
Yeah, this is awesome!! Keep us posted on how it goes!

PR #1930, opened 2024-11-04
Move project metadata from setup.py to pyproject.toml

mwtoews:
Similar to https://github.com/keras-team/keras/pull/20427, this PR migrates the core project metadata into `pyproject.toml` using [PEP 621](https://peps.python.org/pep-0621/), as supported by [setuptools](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html).

Some changed metadata is to use the same name and contact email as `keras` (the former seemed to have a bogus email, not that it's a big deal). It adds some useful URLs.

The resulting *.whl file excludes installing `benchmark` (this is a fix), but includes `integration_tests` (neutral change).


haifeng-jin (approved):
Thanks for the PR!

PR #1931, opened 2024-11-06
Replace Black, isort and flake8 with Ruff check and formatter

mwtoews:
This PR replaces Black, isort and flake8 with [Ruff](https://docs.astral.sh/ruff/).

Changes to the source files are due to `ruff format`; there were no changes required for `ruff check`.

Other changes in this PR:

 - Python version in docker/pre-commit.Dockerfile is upgraded to minimum version
 - Options for pytest and cov are migrated from setup.cfg to pyproject.toml
 - Shebangs and bash options in shell scripts are added, similar to the main keras repo

See also these recent changes to keras:

- https://github.com/keras-team/keras/pull/20442
- https://github.com/keras-team/keras/pull/20445


haifeng-jin (commented):
Thanks for the PR!
We would prefer to keep the linter and formatter unchanged for the moment to avoid any unexpected new workload added by it later.
The current linter and formatter work fine for now.

